PR #105, opened 2015-07-15
Update cfcontent.json

fmdano:
added a tag and script example, script needs verified.


pfreitag:
I don't think the script example is applicable to cfcontent, more applicable to cfsavecontent.

PR #112, opened 2015-07-22
Update cfcase.json

fmdano:
add tag and script examples


pfreitag:
FYI a few of these had JSON errors, the double quotes need to be escaped with a backslash, like this: \"

PR #125, opened 2015-07-24
Update replace.json

fmdano:
added to examples to show no scope and with scope

### DIFF
--- a/data/en/replace.json
+++ b/data/en/replace.json
@@ -20,6 +20,20 @@
 	},
 	"links": [
 
+	],
+	"examples": [
+		{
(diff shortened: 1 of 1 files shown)


pfreitag:
A few notes... you don't need to say Expected Result in the result string that is added automatically. And if it is a function one liner you can just say "code":"Len(\"test\")" and it will change it to: <cfoutput>#Len("test")#</cfoutput> in the trycf editor popup, this keeps the example pretty clean.

pfreitag:
Also the title on the second example is "Tag Syntax" which doesn't apply here.

PR #132, opened 2015-10-02
added link to adobe PDF Service Manager setup/usage guide

chopki:
Hopefully this is the correct way to go about doing this - i dont use PR's internally and have never forked or done a pr to a seperate project.

Either way thought the link i added would be useful on this page as i found it hard to find when having trouble getting cfhtmltopdf set up properly.


pfreitag:
Thanks Chris - there is actually a links section that we can add this to in the JSON structure but this is good for now

PR #138, opened 2015-10-03
Add StructEach Function

maryjos:


pfreitag:
Good one - Thanks!

pfreitag:
I saw you also updated the indexes thanks -- there is a utility that generates them (in utilities folder I think) so you don't have to do it manually when you add something new

maryjos:
Ah okay, thanks, good to know!

maryjos:
Oh I see it included my update to add the new Index page for Struct functions as well. Figured that should be in there since there was one for arrays.

maryjos:
Boy, that leaderboard sure is slow to update though! Still haven't made it on (I would be contributing anyway, tired of waiting for Adobe to accept my help with docs, so will try to make these as good as possible!)

maryjos:
Just noticed something too... you may want to add a secondary sort for number of lines for people tied on number of contributions. Lot of people with 1 contribution in there but seems to be random how they are ordered.

maryjos:
I could of course do the change and check it in but I figure that's something the person in charge of the project should decide to do or not. ;-)

pfreitag:
Yes for some reason you are not showing up on github under the contributors it usually happens after a few minutes. 

Feel free to update the sorting if u want

maryjos:
Ah yeah I looked at what it returns, not easy to do that kind of sort. Probably not worth the trouble.

PR #151, opened 2015-10-07
Add StructFilter Function, Member Support Syntax, Other Updates

maryjos:


maryjos:
Okay, merge conflicts should be fixed now. Didn't realize the forks wasn't kept up to date with the original.

PR #156, opened 2015-10-09
ALLCAPSified abbreviation within function name

stelt:


pfreitag:
Actually we are going for lowerCamelCase, so that one should probably start with lower case `i` as well

PR #157, opened 2015-10-09
changed "IsIs" to "Is" as start of function name

stelt:


pfreitag:
This is an actual function the LS prefix stands for locale specific, https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-l/lsisdate.html

stelt:
I was afraid something like that was going on, thanks. Not very clear. 
Especially with capital I and lowercase l looking the same, making lsIs look as a repetition

PR #158, opened 2015-10-09
changed "IsIs" to "Is" as start of function name

stelt:


pfreitag:
https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-l/lsiscurrency.html

PR #173, opened 2015-10-29
Added full Application.cfc callback function example

infernocloud:


atuttle:
Personally I'd rather not refer to them as "callbacks." https://en.wikipedia.org/wiki/Callback_(computer_programming)

Since we're not passing them as arguments, I think a name that fits better is "application/request lifecycle methods"

infernocloud:
Good point. I always think of them as magical built-in callbacks since they happen in response to events, but they aren't technically callbacks. I've updated the language to reflect that.

pfreitag:
Thanks for the pull request Michael, and @atuttle thanks for pointing that out, I prefer lifecycle methods as well.

PR #175, opened 2015-10-30
Added cfparam examples

infernocloud:


atuttle:
Additional supported syntax:

`param name="foo" default="bar";`

pfreitag:
Merged this - the cfscript example in the PR is CF11+ syntax I think. 

Adams suggested syntax is supported a few versions back. not sure which version (on my phone now) but feel free to add a PR with such an example if you are so inclined.

PR #181, opened 2015-11-17
attempt to fix busted link rendering

atuttle:


atuttle:
@foundeo you said you had tried converting it to an explicit link, but I'm not sure if you tried putting that link on other text, or tried to keep using the url itself as the link text. If this works, great, if not, you are of course free to close without merging.

pfreitag:
Yeah looks like it didn't fix it but it is probably better to have the explicit link syntax so merging anyways

PR #182, opened 2015-11-17
Create canonicalize.json

yogidevbear:


pfreitag:
Hi -- thanks for the pull, however it failed the integration tests: https://travis-ci.org/foundeo/cfdocs/builds/91601554#L153 invalid json.

yogidevbear:
Hi Pete. Just trying to figure out my issue with the Travis CI stuff. I'm under the assumption that I added `canonicalize` to index.json and functions.json, but not in the same branch so to speak which is why I'm seeing the errors in https://travis-ci.org/foundeo/cfdocs/builds/91605233

yogidevbear:
I think the last errors in Travis CI should resolve now (I hope).

yogidevbear:
So one of the errors is https://travis-ci.org/foundeo/cfdocs/builds/91605233#L306, but canonicalize is not a tag function. It appears to be cfscript-only.

yogidevbear:
Error between keyboard and chair :smile: 
I had the wrong type set to tag instead of function

pfreitag:
Ok I will take a look shortly - I'm out of the office right now -- I think the issue is probably that lucee (which is what Travis is running) may not support the function?

pfreitag:
Oh I see test passes now nice job!!

PR #187, opened 2015-11-19
ESAPIEncode

yogidevbear:


pfreitag:
Thanks for the pull -- however there is a conflict in the all.json file... If you want to just send the esapiencode.json I can take care of updating all the indexes.

PR #201, opened 2015-11-24
Modified encrypt.json

ddspringle:
Added example using CBC mode encryption


pfreitag:
Thanks Denard good example!

PR #202, opened 2015-11-24
Add csrf functions

elpete:
Updated indexes as well.


pfreitag:
Well this is a first, I just merged a pull from @ddspringle that also had the CSRF functions -- can you guys compare your work and see if there is anything in this pull that I should merge or should we just keep Denards version intact.

ddspringle:
Great minds think alike! They look about the same... I didn't update the indexes except for security, so maybe grab the updated indexes from this one? My examples seem more relevant and I added additional text to the descriptions, etc to help identify what these are used for - but other than that, looks like we're all on the same page here ;)

elpete:
Sounds good to me. 👍



Cheers,
Eric

pfreitag:
Thanks guys, merged in the index changes.

PR #205, opened 2015-11-24
Documentation for spreadsheetNew

Twillen:
I have also added a rewrite rule for tryCF in IIS, 
Also found a small syntax issue in a test that CF didn't like.


pfreitag:
Thanks!

PR #209, opened 2015-11-25
xmlFormat() and <cfabort> examples

ddspringle:
And a tweak to the example for urlDecode() to use something more relevant than 'myPassword' and to expand on the rendered output to demonstrate what is actually going on, I hope ;)

P.S. Good add of the simple example... hadn't thought of something that simple heh!


ddspringle:
Also added examples for cfcollection

PR #219, opened 2015-12-02
Updated the search functionality to do partial matches and 'search in…

boyzoid:
Updated the search functionality to do partial matches and 'search in string' functionality.

Increased the number of suggestions that will appear in search results.


atuttle:
The description sounds awesome. How about a screenshot showing it in action? @boyzoid you can capture it as you normally would and then just paste it directly into the github comment box. :)

pfreitag:
Thanks Scott -- merged it... The rollout on cfdocs.org will be gradual due to CDN caching

PR #220, opened 2015-12-03
Missing examples to menu and bug fix.

murphydan:
Adding the missing examples page to the "Other" menu and fixing a bug that was preventing reports from showing.


pfreitag:
thank for the bug fix as well!

PR #221, opened 2015-12-03
added some image script examples

bmdoherty:


pfreitag:
Wow - that was an impressive effort, thanks for all the examples!

PR #228, opened 2015-12-05
Examples for few simple functions

cfmitrah:
Examples for few simple functions (arrayavg, arraylen, insert, isarray & len)


pfreitag:
Thanks!

PR #229, opened 2015-12-05
Add example syntax highlighting

ddspringle:


ddspringle:
This adds some (very) basic syntax highlighting to the examples (e.g. highlight 'len' in if( len( someVar) ); etc.) to make it easier to find the tag, script or member function in use within the code

ddspringle:
Right, so, this isn't going to be perfect but it captures most of the syntax needed to be highlighted while eliminating most of the false positives now.

PR #231, opened 2015-12-09
Examples for few simple functions (listToArray, listSort, listQualify)

cfmitrah:
Examples for few simple functions (listToArray, listSort, listQualify)


pfreitag:
Nice work -- thanks

PR #240, opened 2015-12-18
normalize function names

jcberquist:
Function names are now in camel case. The precise capitalization of each function throughout each file has also been matched so it does not vary in the file (hopefully).

This is a rather large pull request. If you need it done in smaller chunks, I am happy to redo. Only function names should be touched in each file, aside from a few trailing spaces my editor removed.


pfreitag:
Thanks!!

PR #242, opened 2015-12-18
updated supported engines for functions

jcberquist:
This mainly involved checking links to ensure they were valid. I should note that all Railo links redirect to the Lucee docs main page, so I couldn't really do much with those. I only removed Lucee/Railo from the engines if they were noted as unsupported in the Lucee docs. So there are still some invalid links present. I didn't remove invalid ColdFusion links either.


pfreitag:
Thanks, great work!

For Railo I am not too concerned since the cfdocs.org site does not actually do anything with the Railo engines data currently (unless there is a note it will display it). I considered deleting it, but decided to keep it in incase anyone wants the metadata.

PR #245, opened 2015-12-21
Examples for few simple functions (arraynew, arraysort, arraysum, ran…

cfmitrah:
Examples for few simple functions (arraynew, arraysort, arraysum, rand, val, structSort)


pfreitag:
Nice work thanks!!

PR #246, opened 2015-12-21
added Lucee specific functions (metadata only)

jcberquist:
Thanks for taking a look. I didn't add any of the functions to the various function lists, or to the index. If you want to proceed, I can withdraw this pull request and create a new one that includes that data, or I can add another pull request that add the functions to the indexes and lists.

### DIFF
--- a/data/en/applicationstarttime.json
+++ b/data/en/applicationstarttime.json
@@ -1,0 +1,14 @@
+{
+	"name": "applicationStartTime",
+	"type": "function",
+	"syntax": "applicationStartTime()",
+	"returns": "date",
+	"related": [],
+	"description": "",
+	"params": [],
+	"engines": {
+		"lucee": {"minimum_version": "", "notes": "", "docs": "http://docs.lucee.org/reference/functions/applicationstarttime.html"}
+	},
+	"examples": [],
+	"links": []
+}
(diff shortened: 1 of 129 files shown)


pfreitag:
Thanks -- I will merge this and then you can do the indexes as another PR...

One question - I noticed a couple odd functions like cfusion_encrypt, cfusion_decrypt (which are the same as passing CFMXcompat to the encrypt/decrypt function and hash40 which is for some sort of backwards compatibility... Do you think these are worth adding? Maybe, not sure.

jcberquist:
I don't feel strongly about those. A quick search for "compat" and "deprecate" in the functions reveals the following list:
```
cachekeyexists
cfusion_decrypt
cfusion_encrypt
componentinfo
getcurrentcontext
hash40
imagedrawimage
listtrim
nowserver
valuearray
```
The only function there that I have used is `valueArray()`. I can remove any or all of these if it seems best.

jcberquist:
With the indexes, which ones are case-sensitive and which not? I would prefer to have casing preserved where possible.

pfreitag:
Just run `/utilities/indexer.cfm` and it will generate the index files for you.

pfreitag:
btw I updated the doc view to highlight when only one engine is supported by a tag or function.

pfreitag:
might be useful to add a category `lucee-specific` though there might be some categories we could generate dynamically instead of manually by using the metadata -- that being one of them.

jcberquist:
Thanks for the info on `indexer.cfm`, that's convenient.

Issue #226, opened 2015-12-04
CFImage tag needs examples and some updates.


pfreitag:
Needs examples resize would be a good first example. 
Needs min_version:8 
Missing Attribute: interpolation CF10+ applies to action = "resize"

mjhagen:
Is anyone working on this one? If not, I'd like to take a stab at it.

pfreitag:
Nope -- no one is working on it, feel free to tackle it

mjhagen:
Alright, submitted PR #253 with basic examples, maybe there should be a recommendation in there that points to the image*() functions?

pfreitag:
Thanks for all the examples...

I added some of the image functions to the `related` key in the json file, and I also linked to the Image Functions category in the links area.

PR #253, opened 2015-12-23
cfimage examples

mjhagen:
Added basic examples for each cfimage action and added the defaults for
the params.


pfreitag:
fixes #226

PR #259, opened 2015-12-28
Fix markdown links

mjhagen:
Markdown links `[text](url)` get screwed up because the autolink method
replaces links that are already links. This should fix it by only
replacing links that don't have `"` in front of them.


pfreitag:
Thanks!

PR #261, opened 2015-12-31
Cfmodule

dhjones64:
Added examples for <cfmodule> tag. Bug fix for unescaped angle brackets in attribute reference section for "name" attribute.


pfreitag:
Thanks Dave!

PR #262, opened 2016-01-04
Updated supported engines for tags

jcberquist:
Lots of formatting changes too, hopefully that is okay. The only data changes are in the engines supported.


pfreitag:
thanks!

PR #268, opened 2016-01-13
Fixing error in run code for StructReduce

nodoherty:
Also added a dump of the original struct in the result for reference.


pfreitag:
Thanks!

PR #274, opened 2016-01-27
deserializeJSON() replace escaped double quotes

mjhagen:
I agree with Adam
(http://blog.adamcameron.me/2016/01/coldfusion-how-not-to-document-function.html)
this makes the example more focused.


pfreitag:
Thanks I agree with @adamcameron here as well, that is much easier to read.

PR #276, opened 2016-01-28
Update cfhttp.json

petrichor8:
typo fix


pfreitag:
Thanks for fixing that David!

PR #279, opened 2016-02-05
CF11+ cfscript function call

oldAndBold:
Add example of script function call and spelled "recipient' correctly


pfreitag:
Thanks for the PR -- it is failing the automated json format tests due to a missing comma in the examples array. Are you able to fix it? If not I'll take care of it - thanks for the example and fixing the misspelling!

pfreitag:
Thanks!!

PR #312, opened 2016-02-22
Fix link to ArrayFindAll

ryandeba:
Link was pointing to "/ArrayFindAt" which doesn't exist. I suspect that was meant to be "/ArrayFindAll"

### DIFF
--- a/guides/en/closures.md
+++ b/guides/en/closures.md
@@ -282,7 +282,7 @@ http://taha-sh.com/blog/understanding-closures-in-javascript
 * [ArrayFilter](/arrayfilter)
 * [StructFilter](/structfilter)
 * [ListFilter](/ListFilter)
-* [ArrayFindAt ](/ArrayFindAt)
+* [ArrayFindAll ](/ArrayFindAll)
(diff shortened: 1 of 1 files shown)


pfreitag:
Thanks I think so, pinging @fmdano as he wrote the guide. Dan can you double check this?

fmdano:
seems like I missed most of the email chain on this...did I not update a link right in the arrayFindall? If there is an issue in the file, just change it...is that what someone wants to do? 
dan

Date: Thu, 25 Feb 2016 12:51:50 -0800

pfreitag:
@fmdano - in your http://cfdocs.org/closures guide there was a mention of an `ArrayFindAt` function (which is not documented by Adobe) in the list of closure functions, @ryandeba changed it to `ArrayFindAll` thinking that was what you had meant.

I just wanted to ping you to make sure we weren't removing a reference to some undocumented function.

fmdano:
Ah I see, no that was a typo...darnit guess i was typing too fast or adding typos to see if others would catch them...LOLglad he caught it...
thanks for letting me know about that.dan

Date: Fri, 26 Feb 2016 07:13:07 -0800

PR #313, opened 2016-02-24
Just making the tag syntax equivalent to the script syntax.

cfvonner:
I see that the full syntax for both is demonstrated in the code examples at the bottom, but the top examples implied to me (at first glance) that the tag version does not have the `statusCode` attribute.


pfreitag:
Good catch, thanks!

PR #314, opened 2016-02-25
Added examples for <cfparam type="regex"...> and IsDefined("form.var_name")

mborn319:
Edited isdefined.json and cfparam.json.


pfreitag:
Thanks for the examples!

PR #315, opened 2016-03-01
htmlEditFormat(): added deprecated and cleanup

mjhagen:
also: fixed a </div> that was in the wrong place on doc.cfm (at the end
of the Compatibility section)


mjhagen:
Fixed, sorry about that I didn't realize until after submitting :/

PR #317, opened 2016-03-01
homepage: responsive fix for contributors section

mjhagen:
A couple of fixes for the homepage, so the contributors section looks
better on tablets (2 col) and phones (1 col).


pfreitag:
Looks good thanks!

mjhagen:
:+1: cool :)

Issue #318, opened 2016-03-02
fileUpload


henrylearn2rock:
I believe, new for CF10, there's a 5th argument for fileUpload() that define if Strict is true (default) or false.

mjhagen:
There is, but I've not found it in the Adobe (or Lucee) docs, do you have more info on it?

henrylearn2rock:
Unfortunately I can't. That's why I hope this doc can include it at least

henrylearn2rock:
Thanks!

Issue #319, opened 2016-03-02
fileUploadAll


jamieruntime:
Just a minor note: the description of the function says "**fileUpload**(string destination,[string accept,[string nameConflict]]) → returns array" instead of **fileUploadAll**

pfreitag:
This has been fixed by @mjhagen thanks Mingo, and thanks @jamieruntime for reporting it!

PR #320, opened 2016-03-02
updates to fileUpload and fileUploadAll

mjhagen:
As mentioned in #319


mjhagen:
Should fix #318

mjhagen:
hmm, after some more testing, the description for `strict` needs some more work.

Issue #299, opened 2016-02-16
CF2016 Updated Application.cfc / cfapplication


pfreitag:
added passArrayByReference and searchImplicitScopes

PR #328, opened 2016-03-10
Application.cfc: passArrayByReference and searchImplicitScopes

mjhagen:
Added descriptions for the new CF2016 application settings. (Issue #299)


pfreitag:
Thanks Mingo - btw you should be able to use a back tick for a <code> tag

mjhagen:
ah, cool, I'll do that from now on.

PR #329, opened 2016-03-10
throw: swapped message and type arguments

mjhagen:
They now reflect the order in which they are used in CF.


pfreitag:
Good catch, thanks

PR #332, opened 2016-03-14
Google code prettify and 2 additions to cf2016.json

mjhagen:
Not 100% satisfied with it, since it doesn't go into javascript mode when encountering a <cfscript> tag, but I think that might be fixable if we're not using the cdn version. I've tried multiple others but feel most are either too slow or like swatting a fly with a sledgehammer. I'm submitting the pr now, so you can have a look at it.


pfreitag:
Thanks @mjhagen, I tested a few pages and it looked pretty good. For the part you are not satisfied for are you referring to the case where the code example is inside a cfscript tag? 

If so I tend to omit the cfscript tag in those cases, the example runner code is smart enough to wrap cfscript tags around an example so it will run... and I think readers can easily know if an example is cfscript or not (hopefully)

Can you point to an example tag / function where it is not working as well?

mjhagen:
It's for the ones where there is a combination of (in this case html) tags and cfscript, like http://cfdocs.org/fileupload In the prettify script is a part that looks for `<script></script>` to switch to javascript highlighting that I figure we could use for `<cfscript>` blocks, but couldn't get that to work yet.

mjhagen:
See https://github.com/google/code-prettify/blob/master/src/prettify.js#L1364

pfreitag:
Yeah -- could fix that to support cfscript pretty easily I think, use something like `/^<[cf]*script\b[^>]*>([\s\S]*?)(<\/[cf]*script\b[^>]*>)/i]`

I am going to merge this PR thanks!

pfreitag:
Oh btw no need to manually update the cf2016.json file, it can be updated by hitting /utilities/indexer.cfm on a local copy (sorry I forgot to do that)

mjhagen:
Hmm, yeah, I tried that (using `(cf)?script`) but couldn't get that to work yet, so figured I'd just submit this now and make it nicer in the future.

pfreitag:
sounds good thanks.

PR #333, opened 2016-03-14
Updated docs for CacheGet

atuttle:


pfreitag:
Thanks Adam -- btw this is one of those cases where Lucee has a different method signature.

http://docs.lucee.org/reference/functions/cacheget.html

I will update the doc with some more info about the incompatibilities.

Issue #308, opened 2016-02-16
CF2016 Updated structNew


pfreitag:
Added argument: structType

PR #335, opened 2016-03-15
structNew and cfqueryparam

mjhagen:
- Added examples and the new structType argument #308 to structNew
- Fixed cfscript example on cfqueryparam


pfreitag:
Thanks for continuing to chip away at the CF2016 enhancements!

PR #344, opened 2016-03-25
Add CF11 additions to CFHTTP.

cfvonner:
Add some new attributes introduced in CF11 (authType, domain, workstation). Add missing attribute multipartType. Add PATCH to method attribute (added in CF11).


pfreitag:
Nice work thanks!

PR #349, opened 2016-04-19
Add ArrayReduce to list of array member functions

ryandeba:
it works - http://trycf.com/gist/24a146715b823f183039dcd5cf0c417a/acf2016


pfreitag:
Thanks!

PR #351, opened 2016-04-20
Fix auto-linking of function names

cfvonner:
Add trailing space after function names so links to individual doc pages can be built automatically


pfreitag:
Thanks!

PR #354, opened 2016-05-03
Fixed typos

martypaz:


pfreitag:
Good catch, thanks!

martypaz:
Hi Pete – I’m in the process of automatically re-creating the DreamWeaver extension for CF11 and was using the JSON from your site to do so. I was wondering if you would like the new extension to be on CFDocs – I know there are still a few devs out there still using DW. I must admit I can’t stand CFBuilder and use the design mode in DW for lots of my work.

 

I’m also going to produce a version for Visual Studio – Mostly as a proof of concept to see if it would be an alternate editor/debugger.

 

Let me know if you want the extension/builder code.

 

Martin

PR #355, opened 2016-05-19
ORM Settings guide

mjhagen:
- added this.ORMSettings guide
- added markdown flag to autoLink() function so `<br>`s are only added
once.


pfreitag:
Thanks!

Issue #356, opened 2016-05-23
structmap and listmap functions missing lucee support


ryanguill:
The following functions say (incorrectly) that they are not supported on lucee

* http://cfdocs.org/structmap ↝ http://docs.lucee.org/reference/functions/structmap.html
* http://cfdocs.org/listmap ↝ http://docs.lucee.org/reference/functions/listmap.html

`querymap` says it isnt supported by ACF but it is, but I think there is already an issue for that.

atuttle:
Sorry for multiple PR's to fix this one issue -- I just edit directly on github and it's easier to do it this way. :)

pfreitag:
No problem @atuttle whatever works for you - thanks for fixing!

ryanguill:
yes, thanks @atuttle! 💯

PR #357, opened 2016-05-23
For #356, fix lucee compat warning for StructMap

atuttle:
I couldn't find any docs on minimum version of Lucee for this function so I just guessed 4.5


pfreitag:
Thanks, looks like they were added to Railo 4.0 so Lucee 4.5 is accurate.

PR #362, opened 2016-06-06
Update cfschedule.json

KrunchMuffin:
Added group and mode attributes


pfreitag:
Thanks @KrunchMuffin!

PR #367, opened 2016-06-13
Update directorylist.json

mjhagen:
Moved some argument notes to the engine section.

### DIFF
--- a/data/en/directorylist.json
+++ b/data/en/directorylist.json
@@ -49,7 +49,7 @@
 		},
 		{
 			"name": "sort",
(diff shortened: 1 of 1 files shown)


pfreitag:
Thanks -- I want to keep the CF9+ on the type attribute so I will put that back in (I find those handy), but thanks for updating the engine notes with more clear info.

pfreitag:
Oh -- disregard that comment, since the function was introduced in CF9 it is not necessary to have that CF9+ on the type attribute. I will keep as is.

PR #368, opened 2016-06-21
Use `spreadsheetcreatesheet` in the example

aliaspooryorik:
Hope the formatting of the example works!

### DIFF
--- a/data/en/spreadsheetcreatesheet.json
+++ b/data/en/spreadsheetcreatesheet.json
@@ -21,7 +21,7 @@
 		{
 			"title":"Tag Example",
 			"description":"The following example creates two sheets: CourseData and EvaluationSheet. ",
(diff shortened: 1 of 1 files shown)


aliaspooryorik:
Build failed due to:
`An error occurred running the tests. Message: [invalid component definition, can't find component [testbox.system.TestBox]]`

ddspringle:
I cannot speak to why that error occurs, but it has happened to me in the past  though it's probable that it is due to a network issue grabbing the box.zip.

On another note, from your commit it looks like you've replaced the tag example with a script example. I would instead add a second example to this function, denoted as a script example, which highlights the additional way of handling generation of spreadsheet data.

pfreitag:
@aliaspooryorik @ddspringle thanks - the Travis builds have been fixed it was due to some testbox/commandbox/forgebox changes.

PR #374, opened 2016-07-13
Addl CSRF examples

ddspringle:
Based on discussion in Slack, added an example of using CSRF with multiple open tabs to the same form


pfreitag:
Thanks!

PR #375, opened 2016-07-18
update cffunction.json

KrunchMuffin:
added REST attributes.

Edit: forgot to add default of "/" for produces and consumes


pfreitag:
Thanks for the update @KrunchMuffin! There was a minor JSON issue but I fixed it.

PR #376, opened 2016-07-22
DateDiff doesn't support the millisecond date part.

laurencehoess:
https://forums.adobe.com/thread/228381


pfreitag:
Thanks Laurence!

PR #378, opened 2016-08-08
Higher-Order-Function Links

ryanguill:
Added some links to many of the higher-order-function taking functions
removed unnecessary escape slashes from json documentation for link urls


pfreitag:
Thanks Ryan, and also thank you for fixing the readme!

PR #382, opened 2016-08-17
Fix case issue with returntype attributes

rossmd:
The available attribute options for returntype in the tag had some of the arguments with capital first letters:

["Any","Array","Binary","boolean","date","guid","Numeric","Query","String","Struct","UUID","variablename","void","xml","(component name)"]

It was odd to me that some had a capital first letter and others did not. All other attribute options are all lowercase.


pfreitag:
Thanks @rossmd!

PR #384, opened 2016-08-18
Responsive html update

mjhagen:
As promised (a long time ago) here's my attempt at a more responsive arguments/attributes table. As well as the tag/function listing.


pfreitag:
Looks good thanks!

PR #390, opened 2016-08-25
New operator

penhorwood:
New operator page under the Other dropdown.  Documents most if not all of the functions of the new operator added in CF9.


pfreitag:
Awesome work thanks!

PR #392, opened 2016-08-26
adding new to dropdown menu

penhorwood:
I forgot to add the new operator doc to the menu.


penhorwood:
I edited the same lines of code so the CI stuff is crashing but the end product should pass CI.  I realized that most of these were trig functions so I added related links  starting with log.

If you want I can cancel the pull request but I am very green with it comes to dealing with GitHub

PR #394, opened 2016-08-27
math function examples

penhorwood:
I figured out the issue on the examples.  The problem is Windows and Linux version give different precision results.  I ran all of my examples on cftry.com but only on Windows.  The CI stuff runs on a Linux box it appears which caused some of the test to fail.

I am trying the pull request again since I can't figure out how to run the test outside of a pull request.


pfreitag:
Thanks @penhorwood!

BTW to run the tests the easiest way would be to use CommandBox to run a local server `box recipe server-start.boxr` ( see the main README file for info more info on that), then once you have a server running you would just hit http://127.0.0.1:8411/tests/run.cfm

penhorwood:
Thanks for the pointer on running test.  I will do that from now on.  I got commandBox and the local copy of cfdocs.org running over the weekend.  The new page is not showing up on the dropdown menu.  I though I had that one fixed as it was my main concern.

pfreitag:
The main cfdocs.org site uses Amazon CloudFront CDN and is setup to do some caching, so it will just take some time for it to start showing up on the site. I can see it in my local copy so all should be good.

Issue #381, opened 2016-08-16
cfwebsocket sub-channel example


sjmjustin:
In the example on sub-channels, the sub-channel applies to the channel being subscribed to (e.g. the subscribeTo attribute or subscribe() JavaScript method).  The example implies that the sub-channel is specified in the name attribute which is not correct (name is the the JavaScript object that will be created).

pfreitag:
Thanks @sjmjustin for reporting this. Can you suggest a better example? If so you can either post it here and I'll update it, or you can submit a pull request which makes it super easy for me.

pfreitag:
Fixed in PR #398

PR #398, opened 2016-09-06
Update cfwebsocket.json

sjmjustin:
Updated subchannel example to use correct subchannel subscription method.


pfreitag:
Thanks! This fixes #381

PR #402, opened 2016-09-14
leaderboard based on effort

penhorwood:
effort = lines altered + ( commits * factor )

currently factor is set to 10.  The leaderboard information is wrapped up into a query so it would be easy to change what data point controlled the sort.


pfreitag:
Cool thanks was thinking about doing some tweaks to that - I might also add something that gives points for recent contributions so you move down a bit if you haven't contributed in a while.

pfreitag:
Just added weeksAgo and show points on leaderboard

PR #403, opened 2016-09-15
Fixed some things things that were breaking tests

mjhagen:
I think some non-parsing json broke the missing-examples page.
Also fixed some examples that weren't going through the test suite.


mjhagen:
Well... isn't that ironic :) pr to fix tests, fails tests... let me see what's going on.

pfreitag:
Thanks!

PR #404, opened 2016-09-16
Correcting expectation, should be toBeTypeOf

zakarym:
Noticed that "toBeOfType" should actually be "toBeTypeOf".


pfreitag:
Thanks!!

PR #405, opened 2016-09-18
Add example for queryColumnList

jeremyhalliwell:


pfreitag:
Thanks Jeremy!! 

The integration tests (Travis CI) are failing because your example didn't have a title. 

Also you set `runnable` to false which means it will not have a Run button next to it - it should probably be true because that example should work fine on trycf.com right? I use runnable=false for examples where trycf.com does not work (for example cffile is blocked).

If you have time to make these changes go ahead and update the PR, otherwise I will fix those two minor things when I have a chance. Either way your contribution is much appreciated, thank you!!

jeremyhalliwell:
OK, Pete I've made those changes, and split the example into 2 runnable units. Thanks for providing cfdocs it's such a useful resource, hope I can help out some more!

pfreitag:
Excellent thanks!!

PR #408, opened 2016-09-26
cleaned up invalid JSON

jcberquist:
- `application-cfc.json` had tab chars in the code example
- `spreadsheetaddcolumn.json` had a carriage return in the code example.
In this case I also cleaned the example up a bit.


pfreitag:
Thanks!

PR #410, opened 2016-09-27
Added date member functions

jcberquist:
@pfreitag: Could you review this? I put information about when the member functions were added in the notes for the engines. I am not sure you want it there, but I wasn't sure if there was a better place to put it.

There are two places where I added additional notes:
- `datepart.json`: the syntax differs between CF and Lucee - I added that to the notes under the Lucee engine. 
- `datediff.json`: here it seems that CF and Lucee behave differently with how they diff the dates. I added a note about it to the Lucee engine, as well as an example that can be run that illustrates the difference.

CF2016 also has the `.setSecond()`, `.setMinute()`, etc. member functions, but since there are no corresponding BIFs I wasn't sure what to do with them.

Let me know if anything needs to be changed.

### DIFF
--- a/data/en/createodbcdate.json
+++ b/data/en/createodbcdate.json
@@ -2,6 +2,7 @@
 	"name":"createODBCDate",
 	"type":"function",
 	"syntax":"createODBCDate(date)",
+	"member":"date.createODBCDate()",
 	"returns":"ODBCDate",
 	"related":["CreateDate","CreateODBCDateTime"],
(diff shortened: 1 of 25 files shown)


pfreitag:
This looks great, bummer about that difference in implementation but thanks for documenting it!!! 

I think we will just have to add a new file for setsecond.json and setminute.json and then just note that they are only available as a member function in the docs.

PR #417, opened 2016-10-15
Add TODO menu item

penhorwood:
This gives the missing parts of the documentation more visibility.  I was shocked to find some functions have no description.

I would also like to add some type of flag to the JSON to show that yes the array is empty but it is because there are no params, related, etc items.

Maybe something like 

params:["-none-"]  for a blank item then in doc.cfm those items could be filtered out.


penhorwood:
This pull request adds a "TO DO" menu item to the main page.  All of the missing "reports" are then moved to that menu item.  Currently that is the missing-examples and the missing-description reports.  A number of the lucee only functions have no description which makes the documentation unhelpful.

There is also a missing-params report but it does not show up on the menu since it was more of a proof-of-concept.  This is where I wanted to add a flag to tell the difference between items that were missing and items that truly should not be there.  Not all functions have params but currently there is no way to tell that given the current JSON layout.

pfreitag:
Thanks -- I would like to avoid adding `params: ["-none-"]` because that would probably screw up some IDE integrations such as the Sublime text plugin. There probably isn't many cases of this and I think we can just log an issue on github if we run into any of them.

PR #420, opened 2016-10-19
added note and example of using callback with replace.

ryanguill:


ryanguill:
not sure why its conflicting, I was on an up-to-date master.  can't find a way to see the conflict either...

ryanguill:
I see now that I hadnt pulled from upstream the way I thought I had and that youve beaten me to these changes by 12 hours! I started them weeks ago but just now thought to finish them.

PR #424, opened 2016-10-23
cleaned up descr of queryGetRow, added min ACF version

ryanguill:


pfreitag:
Thanks for all the improvements Ryan!

Issue #298, opened 2016-02-16
CF2016 Updated replaceList


pfreitag:
added new argument: includeEmptyFields

Issue #294, opened 2016-02-16
CF2016 replaceListNoCase


pfreitag:
https://helpx.adobe.com/content/help/en/coldfusion/cfml-reference/coldfusion-functions/functions-m-r/replacelistnocase.html

PR #425, opened 2016-10-23
updated replacelist arguments and syntax

ryanguill:
https://github.com/foundeo/cfdocs/issues/298

I would also like to add a discouraged flag, or at least a note that says that this function is rarely what you are looking for and commonly misunderstood.  Something that explains that it is a (hard to read) shortcut for doing multiple replace statements and not something like replaceListOccurance.

If you agree (and/or have any recommendations for language) let me know and ill update this PR.

### DIFF
--- a/data/en/queryreduce.json
+++ b/data/en/queryreduce.json
@@ -12,7 +12,8 @@
 		{"name": "initalValue", "description": "", "required": false, "default": "", "type": "object", "values": []}
 	],
 	"engines": {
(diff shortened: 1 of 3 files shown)


ryanguill:
also https://github.com/foundeo/cfdocs/issues/294

pfreitag:
Yeah I don't think misunderstood is a good reason to discourage it, but poor readability may be a good reason. You might say Discouraged: Poor readability, use `str.replace(a,b).replace(c,d)` instead.

I don't have a strong opinion on it either way, perhaps you can see what others think on slack.

PR #439, opened 2016-11-24
Missing keysize argument for generateSecretKey.

CraicOverflow89:
Added information for the keysize argument of this function - checked that it definitely exists for Lucee 4.5 and 5, using trycf.com as well as ACF. Although the coldfusion docs say that this was added for ACF8+ (can we add argument-specific version info to the engines section?).

### DIFF
--- a/data/en/generatesecretkey.json
+++ b/data/en/generatesecretkey.json
@@ -1,12 +1,13 @@
 {
 	"name":"generateSecretKey",
 	"type":"function",
-	"syntax":"generateSecretKey([algorithm])",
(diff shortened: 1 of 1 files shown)


pfreitag:
Thanks!!  

Typically what I do if an argument was added in a specific version is to add `CF8+` to the argument description. You can also add engine specific details to the engine keys but I usually use that if there is a implementation difference (for example hash iterations is implemented differently on ACF and Lucee).

pfreitag:
Awesome, happy to have your contributions

PR #442, opened 2016-12-21
Adding reference to onServerStart function

ianosullivan:
onServerStart page needs to be created though

### DIFF
--- a/data/en/application-cfc.json
+++ b/data/en/application-cfc.json
@@ -3,7 +3,7 @@
 	"type":"component",
 	"syntax":"component { this.name=\"myApp\"; }",
 	"returns":"",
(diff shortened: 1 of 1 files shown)


pfreitag:
Thanks -- let's create the page first and then I'll merge this PR, otherwise it will just result in a 404

ianosullivan:
Can do Pete. So I just create the new file here and send a pull request is it? Not overly familiar with the process. 
Ian

pfreitag:
Yes that is correct I would just copy onRequestStart.json and modify it accordingly.

You can send it in a separate pull request

pfreitag:
Thanks again!

PR #451, opened 2016-12-28
Create onServerStart.json

ianosullivan:
Adding page details for onServerStart

### DIFF
--- a/data/en/onServerStart.json
+++ b/data/en/onServerStart.json
@@ -8,8 +8,7 @@
 	"params": [
 	],
 	"engines": {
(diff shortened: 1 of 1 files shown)


pfreitag:
Thanks -- this is not supported on Lucee correct? If not then we need to remove "lucee" from the engines key to indicate that it is not supported.

ianosullivan:
Pete,
I believe you are correct. I've removed the Lucee reference. I can't find a reference to it anywhere. I can't check the Admin though to be 100% as we don't have it installed anywhere at the moment.
Ian.

pfreitag:
I don't see anything in the admin, thanks for adding this - I'll merge it in.

ianosullivan:
Cool, happy to help

PR #452, opened 2016-12-29
Guides not displaying

penhorwood:
The CGI guide was not displaying due to wrong name in the json.

The textbox guide was not displaying locally because that folder exist but the rewrite rule tried using /textbox as the path.  I renamed the textbox.md file to textbox-guide.md and everything works.


penhorwood:
Is there a problem with this request?  The CGI guide is not working which is why I created this pull request.

pfreitag:
Hi Andrew, the only thing I wasn't crazy about was renaming the testbox guide to testbox-guide... I think I would rather solve that problem in the rewrite rule.

penhorwood:
Ok, I think the guide issues are fixed.  If you ever have an issue with any pull request please let me know so I can fix them.  Love the project.  Best docs for CF.

pfreitag:
Thanks @penhorwood -- yes, sorry for not letting you know sooner why hesitated merging it at first.

PR #455, opened 2017-01-05
Minor param corrections to ReFindNoCase

cfvonner:
De-capitalized string param name to be consistent with other params.  Moved the default of "1" from string param to start param.


pfreitag:
Looks good thanks!

PR #456, opened 2017-01-06
Add description to ReFind

cfvonner:
Also, Adobe's description on their help page doesn't mention that it starts from a specified position.  I added that information (which makes it consistent with ReFindNoCase


pfreitag:
Thanks!

PR #460, opened 2017-02-24
Update on cfftp.json

bferdinandus:
Improved the script example so it doesn't throw an exception then you try it. I tested this on CF11 (update 11)


pfreitag:
Looks good thanks!

PR #461, opened 2017-02-27
Updated cookie code example

ddspringle:
Update code examples for cookies to provide solution for folks using this code under CommandBox.


pfreitag:
Hi Denard,

Just curious I was looking at the Lucee docs for `cacheGet` and the signature there is: `CacheGet( id [, throwWhenNotExist [, cacheName ] ] )`

In your example you have the `cacheName` specified as the second argument. I know sometimes lucee will do type checking become compatible with ACF (eg it may check to see if the second argument is a boolean and if not use that for the cache name). 

http://docs.lucee.org/reference/functions/cacheget.html

ddspringle:
Weird... this is referencing a former commit here for some reason (like, months ago) - but yes, Lucee does type checking in this instance and uses it as the cacheName if it's not a boolean, since the method signature is different than it is for ACF (`CacheGet(id, [region])`)

ddspringle:
Closing pull request, will update and send again later w/o conflicts.

pfreitag:
Thanks @ddspringle

PR #462, opened 2017-03-02
Improvements in file upload example

paulklinkenberg:
First, never upload directly to a web accessible directory. Second, always check the incoming file extension. Checking on mime type alone does not do it, this can be easily spoofed.


pfreitag:
Thanks Paul, couldn't agree more.

Issue #459, opened 2017-02-20
cfinput Missing validate="submitOnce"


mborn319:
Added by MX 7. This is NOT supported by Lucee!
http://help.adobe.com/en_US/ColdFusion/9.0/CFMLRef/WSc3ff6d0ea77859461172e0811cbec22c24-7f51.html

```
<cfform action="?action=submit" method="post" class="fmstyle">
<cfinput type="submit" value="Submit" name="submitbutton" id="submitbutton" validate="SubmitOnce" />
</cfform>
```

pfreitag:
Hi @mborn319 that would be great if you can take care of this. 

Basically you will want to edit this file: https://github.com/foundeo/cfdocs/blob/master/data/en/cfinput.json and add it to the `validate` argument. 

You will also want to add a note to the `engines` key under lucee stating that `validate=submitonce` is not supported on Lucee.

PR #466, opened 2017-03-13
cfinput validate submitonce fixes 459

animex:


pfreitag:
Thanks @animex

pfreitag:
Fixes #459

PR #474, opened 2017-04-11
Query documentation update

abramadams:
Adds lazy and psq parameters.  Also fleshes out dbtype description.


pfreitag:
Thanks Abram!

PR #477, opened 2017-04-13
Add cfloop to "see also" section of cfoutput

sbleon:
Because `cfloop(query="x" group="y")` is the cfscript version of `<cfoutput query="x" group="y">`.


pfreitag:
Makes sense, thanks @sbleon!

PR #481, opened 2017-05-16
cacheRemove fixes

mjhagen:
Started with adding missing properties, but continued with a few other cache related functions.


pfreitag:
Thanks Mingo!

PR #485, opened 2017-05-24
Update valuelist.json

shaedrich:
Better description and added missing example


pfreitag:
Hi @shaedrich thanks for the Pull Request!

It looks like the integration test failed because your update was not valid JSON (you have an extra comma at the end of your examples. Would you be able to update and resubmit?

shaedrich:
I fixed this but there is still an error...

pfreitag:
Thanks - it is still failing the integration text because you have the result set to a value that is not the actual result of running the code - I will take care of that, thanks for contributing!

shaedrich:
Okay, thank you.

PR #487, opened 2017-06-14
Encode brackets in <p> in paragraphFormat docs

hawkrives:
The description on https://cfdocs.org/paragraphformat outputs the `<p>` in the example verbatim onto the page, which leads to an empty pair of parens.

I encoded the angle brackets.


pfreitag:
Thanks @hawkrives

PR #488, opened 2017-06-15
add a few descriptions and example for useful functions

NewReStarter:


pfreitag:
@NewReStarter great job, thanks!

One thing I will note, you mentioned that the empty function for Lucee is deprecated in the description, which is great. There is also a way you can denote that a tag/function is deprecated in the `engines` section of the file -- see `ParameterExists` as an example.

PR #496, opened 2017-07-13
Overwrite forced uppercase of functions in listings

richardherbert:
I find reading the listing of functions when forced to uppercase hard to read. I've overwritten the Bootstrap CSS to removed the transformation.


pfreitag:
Thanks for all the PRs, it will take some time for the update to roll out due to caching.

Issue #489, opened 2017-06-21
Param values in JSON


KamasamaK:
What is the meaning of the `values` array in the `params` objects? I assume it is to list the enumerated valid values for instances where those exist. However, I have seen it also used to just give examples.

A good example of this is the [dateDiff function](https://cfdocs.org/datediff). It lists all of the valid values for the `datepart`. It also lists `now()` as a value for both `date1` and `date2`.

ddspringle:
@pfreitag can comment on this more, but from my perspective the `values` array for `params` should only ever be used to enumerate possible values for params that expect them, and not to provide example values.

pfreitag:
Yes it _should_ be used to list the valid possible values when there is a finite list of valid values for an attribute or argument. For example with the `replace()` function the `scope` can be either `all` or `one` nothing else. 

Practically speaking the `values` array is also used to populate autocomplete suggestions for IDE's such as Sublime Text. That is probably why `now()` ended up in there, because it is handy... Technically it probably should be removed.

KamasamaK:
Thanks for confirming this. I don't think its use in autocompletion should influence its use in docs. You could add another key to the JSON for that purpose if it's seen as value. Alternatively, they could look at the param's `type` and just always provide `now()` as an autocomplete suggestion for `DateTime`.

pfreitag:
Yes, I agree - if you want to remove `now()` from `dateDiff` feel free to send a PR and I'll approve it.

KamasamaK:
For this use case, there now exists a property on the param called `suggestions`.

PR #497, opened 2017-07-15
Fixed values for params in date functions. Also improved related.

KamasamaK:
As discussed in Issue #489 I have removed `now()` from `values` anywhere that it appeared.


pfreitag:
Thanks!

Issue #491, opened 2017-06-21
Display types for params


KamasamaK:
I see that `type` is available in the JSON for each `param` but not used in the web view. I am going to take the stance that it should be displayed and ask if anyone has any suggestions on how they think it should be styled and positioned.

pfreitag:
I would suggest styling it with the class `text-muted` which should give it a light gray color, and placing it in the attribute / argument title bar.

PR #506, opened 2017-08-17
Display types for params #491

shaedrich:
Refers to: #491


pfreitag:
Thanks @shaedrich fixes #491

shaedrich:
You're welcome @pfreitag

PR #507, opened 2017-08-17
Added result for ReplaceList

shaedrich:


pfreitag:
Thank you this improves the example -- however I think it could be further improved by just making the entire example this:

`replaceList("My Test String", "Test,String", "Replaced,Sentence")`

Any chance you would like to update that as well?

shaedrich:
That would make it much more clear. Additionally there would be no more unnecessary variables anymore.

pfreitag:
Thanks FYI for future reference you can actually omit the `<cfoutput>` cfdocs will add that in automatically.

shaedrich:
Okay, cool (Y) Thanks for that hint.

PR #508, opened 2017-08-17
Values list header should start in newline

shaedrich:
Was appended to description if provided


pfreitag:
Hi @shaedrich I just updated `doc.cfm` to replace the `xmlFormat` and `htmlEditFormat` with `encodeForHTML` so it is hard for me to see the change you made in the diff. Can you make the change again using the current version of doc.cfm

shaedrich:
Oh, I fixed that (don't mind all these commits :D). But now you should see my changes @pfreitag

PR #523, opened 2017-08-31
Accepts filepath OR fileObject (not just a path)

chopki:
Unsure if there is specific syntax for an argument that can accept multiple different things - See CF docs https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-e-g/filewrite.html


shaedrich:
Do you mean something like `any`?

chopki:
oh no i meant in the cfdocs JSON for explaining multiple different options for one argument e.g an array of possible values for a parameter or something instead of just "this or that".

data/en/filewrite.json
```diff
@@ -6,7 +6,7 @@
 	"related":[],
 	"description":" If you specify a file path, writes the entire content to the specified on-disk or in-memory file. If you specify a file object, writes text or binary data to the file object.",
 	"params": [
-		{"name":"filePath","description":"No Help Available","required":true,"default":"","type":"String","values":[]},
+		{"name":"filePath or fileObject","description":"No Help Available","required":true,"default":"","type":"String","values":[]},
 		{"name":"data","description":"No Help Available","required":true,"default":"","type":"String","values":[]},
```
shaedrich: "No Help Available" should be solved

shaedrich:
Should there be rather two
```json
	"syntax":"fileWrite(filePath, data [, charset])"
```
and
```json
	"syntax":"fileWrite(fileObject, data [, charset])"
```
with explanation in param description of both params?

data/en/filewrite.json
pfreitag: What if we just call the argument `file` and then in the description note that it can be a file path string or a file object, and set the `type` to `any`, for example:

`{"name":"file","description":"A file object or a file system path string.","required":true,"default":"","type":"any","values":[]},`

pfreitag: You would also want to update the syntax to be:

`"syntax":"fileWrite(file, data [, charset])"`

pfreitag:
Just adding my thoughts - I don't think Adobe's doc makes sense here from a CFML perspective, it probably does from a Java Developers prospective where they have two overloaded methods. So they are thinking like this:

`public void fileWrite(java.io.File fileObject, Object data, String charset) { ... }`
`public void fileWrite(String filePath, Object data, String charset) { ... }`

As a CFML developer we would have to implement such code as:

`public function fileWrite(any file, any data, string charset) { ... }`

So I think one signature makes sense with type `any` since although a `fileObject` has an underlying type in CFML, it is not really a type we can satisfy in a `cfargument` or `cfparam` we would have to use `any`. So I don't think having two syntax makes sense here.

data/en/filewrite.json
shaedrich: That sounds good to me :+1:

PR #531, opened 2017-09-05
Clarified usage of 'startrow' and 'maxrows' attributes and added an example for their usage

SebastianZ:
This extends the descriptions of the `startrow` and `maxrows` attributes and adds and example related to them to make it clear how they are used.


pfreitag:
Thank You!

PR #532, opened 2017-09-05
Example with multiple rows as an array in queryAddRow

shaedrich:


pfreitag:
Thanks! FYI wrapping with `<cfscript>` is not necessary in the example, it will assume cfscript if there are no `<cf` patterns in the code.

shaedrich:
You are welcome! I remember, you already told me that. Silly me :D

PR #534, opened 2017-09-06
add tag and script examples for cftimer

notthemonkee:
I wasn't able to get these examples to run properly on the Adobe servers in the pop up runner so I set the runnable property to false to avoid confusion. Let me know if I'm missing something there.

Thanks!

### DIFF
--- a/data/en/cftimer.json
+++ b/data/en/cftimer.json
@@ -18,6 +18,21 @@
 	},
 	"links": [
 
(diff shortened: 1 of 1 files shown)


pfreitag:
Thanks! I think you need debugging enabled for cftimer to show output, so that would probably not work on trycf - runnable:false makes sense, thanks.

PR #538, opened 2017-09-09
JSON schema

KamasamaK:
The JSON schema may need to be tweaked, but I think this is a good basis. The `cfinput.json` file was also modified because it was invalid JSON, but I also made some type changes to match the casing in the schema enums.


pfreitag:
Awesome!!!

PR #539, opened 2017-09-23
directoryList type attribute only supported in Lucee5+

cfsimplicity:


pfreitag:
Thanks @cfsimplicity!

PR #546, opened 2017-10-02
URLEncode - Add Information and examples

Xcreen:
Added various information and examples for URLEncode.

### DIFF
--- a/data/en/urlencode.json
+++ b/data/en/urlencode.json
@@ -2,6 +2,7 @@
 	"name": "URLEncode",
 	"type": "function",
 	"syntax": "URLEncode(string [, charset] [, force])",
+	"member": "string.URLEncode([charset] [, force])",
 	"returns": "String",
 	"related": [],
 	"description": "URL-Encode a String.",
--- a/data/en/urlencode.json
+++ b/data/en/urlencode.json
@@ -18,15 +18,15 @@
 		{
 			"title":"URLEncode() example",
 			"description":"Shows how it url-encode a string.",
-			"code":"URLEncode(\"URL-Encode this string, see also chars like: %&/().\")",
-			"result":"URL-Encode+this+string%2C+see+also+chars+like%3A+%25%26%2F%28%29.",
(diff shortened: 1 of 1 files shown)


data/en/urlencode.json
```diff
@@ -1,18 +1,34 @@
...
+			"title":"URLEncode() example",
+			"description":"Shows how it url-encode a string.",
+			"code":"URLEncode(\"URL-Encode this string, see also chars like: %&/().\")",
+			"result":"URL-Encode+this+string%2C+see+also+chars+like%3A+%25%26%2F%28%29.",
+			"runnable":true
+		},
+		{
+			"title":"URLEncode() usage as a member function",
+			"description":"Same Example like above, but this time we use it as a member-function.",
+			"code":"\"URL-Encode this string, see also chars like: %&/().\".URLEncode()",
```
pfreitag: Can you change the to just be this:

`\"URL-Encode this string\".URLEncode()"`

This example had the same issue as the stringLen example, but the workaround I did for that allowing it to pass. Because this string has special characters in it it is still failing, I'd rather just change the example string than change the test in this case, as I think changing the test might allow it to ignore things it should not ignore.

data/en/urlencode.json
```diff
@@ -1,18 +1,34 @@
...
+			"result":"URL-Encode+this+string%2C+see+also+chars+like%3A+%25%26%2F%28%29.",
```
pfreitag: This would then be:

`"result":"URL-Encode+this+string",`

shaedrich:
@Xcreen Why did you close your pr?

Xcreen:
@shaedrich You already added examples for it (https://github.com/foundeo/cfdocs/pull/558) and also it seem that this will not pass by travis.

shaedrich:
Oh, sorry. I did not have in mind that your pr also added examples. I did not want to pretermit your pr...

Xcreen:
@shaedrich No Problem ;)

shaedrich:
;)

PR #547, opened 2017-10-02
Add examples for ToBase64

Xcreen:
Add String and Object-Binary Example for ToBase64


pfreitag:
Thanks! FYI for the image one you might be able to load an image from a URL to make it runnable.

PR #549, opened 2017-10-02
StringLen - Add Description and Examples

Xcreen:
Add Description and Examples

### DIFF
--- a/data/en/stringlen.json
+++ b/data/en/stringlen.json
@@ -16,14 +16,14 @@
 		{
 			"title":"StringLen() example",
 			"description":"Output the length of the string \"Hello World\".",
-			"code":"WriteOutput(StringLen(\"Hello World\"))",
+			"code":"StringLen(\"Hello World\")",
 			"result":"11",
 			"runnable":true
 		},
 		{
 			"title":"StringLen() using as a member function",
 			"description":"Output the length of the string \"Hello World\".",
-			"code":"WriteOutput(\"Hello World\".len())",
+			"code":"\"Hello World\".len()",
 			"result":"11",
 			"runnable":true
 		}
--- a/data/en/stringlen.json
+++ b/data/en/stringlen.json
@@ -5,13 +5,28 @@
 	"member": "string.len()",
 	"returns": "number",
 	"related": [],
-	"description": "",
+	"description": "Get the length of a string.",
 	"params": [
 		{"name": "string", "description": "", "required": true, "default": "", "type": "string", "values": []}
 	],
 	"engines": {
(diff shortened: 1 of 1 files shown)


shaedrich:
> We like one liner code examples, they can be tested using the unit tests if a result is given. Just omit `cfset` (https://cfdocs.org/utilities/json/)

Just write it right into `writeOutput`.

data/en/stringlen.json
```diff
@@ -5,13 +5,28 @@
...
 	],
 	"engines": {
 		"lucee": {"minimum_version": "", "notes": "", "docs": "http://docs.lucee.org/reference/functions/stringlen.html"}
 	},
-	"examples": [],
+	"examples": [
+		{
+			"title":"StringLen() example",
+			"description":"Output the length of the string \"Hello World\".",
+			"code":"strLength = StringLen(\"Hello World\");\nWriteOutput(strLength);",
```
shaedrich: > We like one liner code examples, they can be tested using the unit tests if a result is given. Just omit `cfset` (https://cfdocs.org/utilities/json/)

Just write it right into `writeOutput`.

data/en/stringlen.json
```diff
@@ -5,13 +5,28 @@
...
+			"result":"11",
+			"runnable":true
+		},
+		{
+			"title":"StringLen() using as a member function",
+			"description":"Output the length of the string \"Hello World\".",
+			"code":"helloWorldString = \"Hello World\";\nWriteOutput(helloWorldString.len());",
```
shaedrich: same suggestion as above.

Xcreen:
@shaedrich Can you review again? Its also failing now due:
> Syntax Error, invalid Expression ["Hello World".len()]

But the example its running fine under Lucee 4.5 / 5

shaedrich:
As far as I can see as well as you this should work. I wonder, too...

pfreitag:
For some reason the `evaluate` function doesn't like it...

The one liner examples are run through evaluate to test the result, I just tried this in commandbox REPL as well, and throws the same error:

```
CFSCRIPT-REPL: evaluate("""Hello World"".len()");
ERROR
Syntax Error, invalid Expression ["Hello World".len()]
```

Not sure how to work around this, since it is an issue with evaluate on Lucee.

pfreitag:
This works `toString('Hello World').len()` not as elegant though

shaedrich:
Couldn't you try to catch this inside the build job, @pfreitag?

pfreitag:
yeah, good call - lets do that I'll work on a fix

pfreitag:
Thanks @Xcreen I updated the test runner to ignore that so it passes now.

Xcreen:
Perfect :)

shaedrich:
Thanks @pfreitag :) :+1:

PR #552, opened 2017-10-03
directoryCreate - added cfscript example

pmturner:
Added basic example of creating a directory after checking if it exists.


pfreitag:
Thanks!

Issue #540, opened 2017-09-28
There are a bunch of functions missing a description


pfreitag:
Go here for list: https://cfdocs.org/reports/missing-descriptions.cfm

shaedrich:
I opened a pull request: #557

PR #563, opened 2017-10-06
Added descriptions for arrayreverse and arraymid

rsericksen:
Made progress on issue #540 

Will add more when I can.


pfreitag:
Thanks!

PR #565, opened 2017-10-06
Optional params

shaedrich:


data/en/replace.json
```diff
@@ -22,18 +22,16 @@
...
 			"description": "",
-			"code": "replace(\"ColdFUsion\", \"U\", \"u\")",
-			"result": "ColdFusion",
-			"runnable":true
+			"code": "getVal = replace(\"ColdFUsion\", \"U\", \"u\"); \r\n writeDump(getVal);",
+			"result": "Expected Result: ColdFusion"
 		},
         {
-            "title": "Example replaing one instance",
+            "title": "Tag Syntax",
```
pfreitag: How is this Tag Syntax?

I actually prefer the more descriptive titles over "Script Syntax" or "Tag Syntax", not that the current titles were very descriptive

shaedrich: @pfreitag Fixed it

Issue #571, opened 2017-10-08
Version validation


KamasamaK:
Would it be appropriate to require a certain format for versions? Specifically the `minimum_version`, `deprecated`, and `removed` within `engines` would be nice to have standardized. It could be a per-engine standard if that seems more appropriate. I'm not sure if there is any validation done on those fields currently, but the JSON schema could be made to validate those against a regular expression.

shaedrich:
There are different formats? Can you tell us some examples and which of them you would prefer?

KamasamaK:
@shaedrich I've changed the title to be more clear. What I'm referring to is if it uses something like Semantic Versioning or its own version format.

For example, [here are release notes for ColdFusion 2016 Update 3](https://helpx.adobe.com/coldfusion/kb/coldfusion-2016-update-3.html). They do add/modify tags and functions in updates as you can see with the introduction of the `StructToSorted` function. Their formatting for the exact version appears to be __2016,0,03,300466__, and I'm wondering if that format should be used exactly. We would want a consistent format/pattern we could validate against.

shaedrich:
@KamasamaK omg. I wonder as well. This looks awful...

I prefer semver but in the end @pfreitag decides which version format to use.

KamasamaK:
Agreed on preferring semver, and it would be simple to convert that to semver format even though it doesn't adhere to the semver semantics.

shaedrich:
Would be easier to handle and read but I don't know if that would work properly...

pfreitag:
Yeah - semver would be best `major.minor.patch` - I don't think we need the additional fourth number (build). In the rare case that you did actually need to indicate which build number you could do it like this: `2016.0.3-build300466` and that would still be valid _semver_

pfreitag:
All that being said, in most cases the version numbers are just the major version so I would want to keep that as valid (no need to add `.0.0`)

shaedrich:
Shouldn't this issue have the label `enhancement`?

KamasamaK:
@pfreitag You're right, build would not be necessary and is optional in SemVer anyway. I suppose the pattern could make `minor` and `patch` optional as well with the `.0`s implied.

PR #575, opened 2017-10-09
Version validation #571

KamasamaK:
Added pattern to JSON schema to validation format of engine version


pfreitag:
Thanks - I am planning on adding a JSON schema check to the travis builds, but not failing it until we can make sure all existing docs are conforming.

PR #581, opened 2017-10-10
Lucee5 new functions and tags WIP

### LINKED ISSUE #341: Lucee 5?

shaedrich:
Resolve #341
- [x] cfloop (times attribute)
- [x] cffile (cachedwithin attribute)
- [x] cfhttp (cachedwithin attribute)
- [x] render
- [x] getTagData
- [x] getFunctionData
- [x] getLocaleInfo
- [x] bundleInfo
- [x] manifestRead
- [x] createObject
- [x] webserviceNew
- [x] cacheregionexists
- [x] cacheregionnew
- [x] cacheregionremove
- [x] GeneratePBKDFKey (related to [GenerateSecretKey](http://docs.lucee.org/reference/functions/generatesecretkey.html), link: [PBKDF2](https://en.wikipedia.org/wiki/PBKDF2))


data/en/render.json
```diff
@@ -0,0 +1,27 @@
...
+	"script":"",
+	"returns":"void",
+	"related":["evaluate"],
+	"description":"Executes CFML/Lucee code that you provide as a string.",
+	"params": [
+	    {"name":"CFML","description":"String to evaluate including some CFML/Lucee code","required":true,"default":"","type":"string","values":[""]},
+ 		{"name": "dialect", "description": "The dialect you want details for", "required": false, "default": "current template's dialect", "type": "string", "values": ["CFML","Lucee"]}
+	],
+	"engines": {
+        "lucee": {"minimum_version":"5+", "notes":"", "docs":"http://docs.lucee.org/reference/functions/render.html"}
```
KamasamaK: Version should not have a `+`

data/en/getlocaleinfo.json
```diff
@@ -0,0 +1,33 @@
+{
+	"name":"GetLocaleInfo",
```
KamasamaK: `name` should start with a lowercase letter

shaedrich:
When I try `generatePBKDFKey('PBKDF2WithSHA512','top_secret','sodium_chloride',2000,8);` on trycf there is always an error. Any ideas why this happens?

> key [DOCUMENTATION] doesn't exist on line 4

Xcreen:
@shaedrich I guess "PBKDF2WithSHA1" - "PBKDF2WithSHA512" is not supported under lucee (on Coldfusion 11+ it work).  
For me (Lucee 5 local) only "PBKDF2WithHmacSHA1" - PBKDF2WithHmacSHA512" is working.

shaedrich:
So, that means that Lucee did not document it good enough...

data/en/generatepbkdfkey.json
```diff
@@ -1,25 +1,25 @@
...
+        {"name":"keysize", "description":"The length in bytes of the key to generate", "required":false, "default":"", "type":"numeric", "values":["true", "false"]}
+    ],
 	"engines":{
-		"coldfusion":{"minimum_version":"11","docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-e-g/generatepbkdfkey.html","notes":""},
-		"lucee":{"minimum_version":"5","docs":"http://docs.lucee.org/reference/functions/generatepbkdfkey.html","notes":""}
+		"coldfusion":{"minimum_version":"11","docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-e-g/generatepbkdfkey.html","notes":"These algorithms are available only in enterprise versions:\n\nPBKDF2WithSHA1\nPBKDF2WithSHA224\nPBKDF2WithSHA256\nPBKDF2WithSHA384\nPBKDF2WithSHA512\nPBKDF2WithSHA512-224\nPBKDF2WithSHA512-256"},
+		"lucee":{"minimum_version":"5","docs":"http://docs.lucee.org/reference/functions/generatepbkdfkey.html","notes":"Lucee seems to support only \"PBKDF2WithHmacSHA1\" and \"PBKDF2WithHmacSHA512\" are allowed values for algorithm argument"}
```
Xcreen: I mean all PBKDF2WithHmacSHA are working from 1 to 512 (so also 224, 256 etc).
But all PBKDF2WithSHA are not working.

shaedrich: In Lucee or Adobe ColdFusion?

shaedrich: According to your previous comment, I guess it's Lucee^^

pfreitag: @Xcreen @shaedrich Adobe ColdFusion Enterprise includes a java crypto provider that implements these algorithms, for Lucee it is up to the provider that you have installed, if using the default java crypto provider it only supports `PBKDF2WithHmacSHA1` on Java 1.7 for example. If you are using Java 8 it supports more algorithms. 

So it is not that Lucee does not support the algorithms, support depends on what crypto provider you have.

shaedrich: Okay, I add this.

pfreitag: Looking at it further, I see you were talking about the non HMAC version of the algorithms. Still it is up to the provider, but it seams the default JCE does not implement the non hmac algorithms at this time. We should add `PBKDF2WithHmacSHA256` `PBKDF2WithHmacSHA512` `PBKDF2WithHmacSHA384` `PBKDF2WithHmacSHA224` and to the algorithm `values`.

pfreitag: It would be good to add something like this to the examples, this will list all valid PBKDF Algorithms on your current engine:

```
for (provider in createObject("java", "java.security.Security").getProviders()) {
  for (service in provider.getServices().toArray()) {
    if (service.getAlgorithm() contains "PBKDF") {
      writeOutput("<h2>#service.getAlgorithm()#</h2><p>#service.toString()#</p>");
    }
  }
}
```

https://trycf.com/gist/5d3f6309803b115f74388b0ef7579b75/lucee5

shaedrich: In my opinion, your function is not part of the doc for that specific function. Adding it to a guide would be better, I guess.

So you can add the non HMAC version in Adobe ColdFision as well? In that case there would not be the need for the enterprise edition.

shaedrich:
I have no clue why this happens...
    
	-> Failure: render.json example result is:hello universe! but evaluated to:. Expected [hello universe!] Actual []

in
https://github.com/foundeo/cfdocs/blob/85ab5dd4b2507d9b9c4d98efba5ba919eee8ad13/data/en/render.json#L18-L26

on trycf everything works well...

KamasamaK:
@shaedrich I believe it evaluates the expression in `code` and compares what is returned with `result`. The `render` function appears to return the empty string on this example. You say it works on TryCF, but you're probably just looking at what is output to the page.

Here is an example that shows what is returned by `render`: https://trycf.com/gist/758a76cdb0cf6c356cc682a84d195271/lucee5?theme=monokai

shaedrich:
But why does this work?

https://trycf.com/gist/b96fb029bbc3542f35b79de7f6858e57/lucee5?theme=monokai

KamasamaK:
@shaedrich The render function outputs the text to the page when executed. So in your example, it is outputting that text and then the `cfoutput` is writing the empty string to the page. It works just as well in [this example](https://trycf.com/gist/b85f8a1fd5da043fc730f6625501ef5e/lucee5?theme=monokai) with no `cfoutput`.

shaedrich:
Ah, okay. It returns back to nothing...

shaedrich:
I also [occured to me before](https://github.com/foundeo/cfdocs/pull/581#pullrequestreview-68557030) that checks fail because a file is not indexed. I thought that is done by the [indexer](/foundeo/cfdocs/utilities/indexer.cfm) automatically...

    -> Failure: getLocaleInfo was not in functions.json index
    -> Failure: getLocaleInfo was not in index.json function index

shaedrich:
@pfreitag For me everything looks fine expect this error message from which I guess the index files should be generated automatically, shouldn't they?

pfreitag:
No the index files need to be updated manually, but you can just hit the `/utilities/indexer.cfm` url locally and it will generate them. It's not ideal but it lets the files be in the repository which means that other tools can use them without having to generate them.

shaedrich:
Okay, thanks. I will do so.

shaedrich:
Unfortunately if I run `box server start` I get this error:

> java.lang.reflect.InvocationTargetException

data/en/generatepbkdfkey.json
```diff
@@ -1,31 +1,37 @@
 {
 	"name":"generatePBKDFKey",
-	"syntax":"generatePBKDFKey(Algorithm, Passphrase, Salt, Iterations, KeySize)",
+	"syntax":"generatePBKDFKey(algorithm,passphrase,salt);",
```
pfreitag: I saw that you removed `keySize` and `iterations` from the syntax... so I just tested to see if they were optional... They appear to be optional on Lucee but Required on ACF, we should probably keep them in the syntax. Also the param `required` is incorrect below, we should probably say they are required but note that they are optional on lucee.

data/en/generatepbkdfkey.json
```diff
@@ -1,20 +1,20 @@
 {
 	"name":"generatePBKDFKey",
-	"syntax":"generatePBKDFKey(algorithm,passphrase,salt);",
+	"syntax":"generatePBKDFKey(algorithm, passphrase, salt, oterations, keySize);",
```
pfreitag: typo: `iterations`

pfreitag:
Thanks for all your hard work on this, looks great - after that iterations typo is fixed I'll merge

shaedrich:
You are welcome :)

KamasamaK:
`getlocalecounty` was incorrectly added to `functions.json` and `index.json` in 6cd559a and bad2415. The correctly spelled `getlocalecountry` was added later, but the misspelled one was never removed.

shaedrich:
Oh, thank's. Must definitely been an instance of scatterbrained behaviour^^

PR #584, opened 2017-10-11
Add description and example to XMLSearch

shaedrich:


data/en/xmlsearch.json
```diff
@@ -4,20 +4,27 @@
...
 	},
-	"links": [
-
-	]
-
+	"links": [{
+			"title":"Title of a blog entry that has good info about this.",
+			"description":"Description of the link",
+			"url":"http://www.example.com/a/b.cfm"
+	}],
```
pfreitag: The link above is an example, did you mean to update that to something else?

shaedrich: I wonder, cause I thought I did.

PR #586, opened 2017-10-13
Fixing invalid or missing return types

KamasamaK:
This should be the last commit for return types. There are also a few other miscellaneous changes done while going through those files, such as spelling corrections or adding `minimum_version`.


pfreitag:
Excellent work - thank you!

PR #592, opened 2017-10-16
Cleanup JSON File Documentation tab indents

shaedrich:


pfreitag:
Thanks - I noticed that the other day too.

shaedrich:
In CONTRIBUTING.md we say "please use tabs", the contributer copied the example and then he had code with spaces that did not correctly match with the tabs he used correctly - poor user :D

PR #594, opened 2017-10-16
Fix wrong charset encoding mask in cfpod

shaedrich:


data/en/cfpod.json
```diff
@@ -5,7 +5,7 @@
 	"returns": "",
 	"related": [],
 	"description": "Creates a pod, an area of the browser window or layout \n area with an optional title bar and body that contains \n display elements.",
-	"discouraged": "The use of tags generating UI is generally discouraged by the CFML community. See: http://static.raymondcamden.com/cfuitherightway/cfpod/index.html",
+	"discouraged": "The use of tags generating UI is generally discouraged by the CFML community. See: <a href=\"http://static.raymondcamden.com/cfuitherightway/cfpod/index.html\">http://static.raymondcamden.com/cfuitherightway/cfpod/index.html</a>",
```
pfreitag: I would rather not have HTML in the document, as we do not know that HTML will always be used to render it.

shaedrich: @pfreitag Okay, fixed it.

shaedrich: Could we run `autoLink()` on it?

pfreitag:
Thanks!

shaedrich:
You're welcome!

PR #599, opened 2017-10-17
Added Examples/Description for ToNumeric

Xcreen:
- Added Description
- Added Default and Values for Radix
- Added two Examples (String/Hex/Bin Cast)


shaedrich:
Value "2-36" for `Radix` argument looks like one string but is a range of numbers. That might be confusing....

Xcreen:
Mhhh thats correct, but have you any suggest?    
Writing every number, also makes no sense...

shaedrich:
That does not make sense either, you're right. What about `"2","...","36"`?

Xcreen:
Mhh you mean: `"2 ... 36"`  ?
Because `"2","...","36"` look also weird due the line-break, than i look like "..." is a own value.

shaedrich:
I used "..." when I wanted to indicate "and so on" so anything after "..." then is to limit that range.

"..." in a single line looks a bit weird but more clear to me because every value has it's own line.

pfreitag:
Thanks @Xcreen

PR #602, opened 2017-10-17
[GUIDE] Date time cheatsheet

shaedrich:


guides/en/dateTime.md
```diff
@@ -0,0 +1,43 @@
+# Date Time Cheatsheet
+## Formatting
+
+| Format | Notes                                                              |
+|--------|--------------------------------------------------------------------|
+| d      | Day of the month as digits; no leading zero for single-digit days. |
+| dd     | Day of the month as digits; leading zero for single-digit days.    |
+| EEE    | Day of the week as a three-letter abbreviation.                    |
+| EEEE   | Day of the week as its full name.                                  |
+| m      | Month as digits; no leading zero for single-digit months.          |
```
pfreitag: So --  the `m` means minutes in `timeFormat`, but month in `dateTimeFormat` that should be clarified and also check to see if there are any other conflicting mask names between the functions.

pfreitag:
Can we lowercase the filename to `datetime.md` 

I think it would be good if the guide was ALL about working with dates and times, so you might show an example of using `<cfqueryparam cfsqltype="cf_sql_timestamp">` etc.

guides/en/datetime.md
```diff
@@ -7,7 +7,7 @@
 | dd     | Day of the month as digits; leading zero for single-digit days.    |
 | EEE    | Day of the week as a three-letter abbreviation.                    |
 | EEEE   | Day of the week as its full name.                                  |
-| m      | Month as digits; no leading zero for single-digit months.          |
+| m      | Month as digits; no leading zero for single-digit months.<br />**CAUTION:** In `timeFormat()` "m" means "minutes"  |
```
pfreitag: Can't use HTML in markdown

shaedrich: Don't you parse it like [GitHub does](https://gist.github.com/shaunlebron/746476e6e7a4d698b373)?

Does your parsing engine allows line breaks anyway?

pfreitag: Uses a java markdown engine... not sure how it handles it but either way, I'd rather not have HTML in markdown. You can test it out by spinning up a local server `box server start` and then browse to the guide - might want to make sure that it is supporting the markdown table syntax.

shaedrich: I tried CommandBox on Ubuntu 16.4 but `box server start` command responded it has been started under that IP but there was no server.

pfreitag: should start on port 8411, try: http://127.0.0.1:8411/

PR #610, opened 2017-10-18
[GUIDE] Scopes

shaedrich:


guides/en/scopes.md
```diff
@@ -0,0 +1,32 @@
+# Scopes
+Scopes are regions of a programm or application in which a variable exists. Variables with the same name in different scopes are different variables.
+
+## Variables Scope
+see Variables Scope
+
+## CGI Scope
+see CGI Scope
```
pfreitag: Add a link to cgi scope guide.

shaedrich: Won't that be done by `autoLink()` function?

shaedrich: It will.

PR #612, opened 2017-10-18
Example simplification & difference clarification

shaedrich:


pfreitag:
very good, thanks!

shaedrich:
Anytime.

Issue #616, opened 2017-10-19
autoLink() and formatting in notes/discouraged param


shaedrich:
`autoLink` should convert strings to proper links in `notes`/`discouraged` param and also format strings to code and other basic formatting as used in [`cacheGet()`](https://cfdocs.org/cacheget)'s Lucee notes for example.

PR #618, opened 2017-10-19
Add Lucee param throwWhenNotExist

shaedrich:
and made script runnable

### DIFF
--- a/data/en/cacheget.json
+++ b/data/en/cacheget.json
@@ -13,7 +13,7 @@
 	],
 	"engines": {
 		"coldfusion": {"minimum_version":"9", "notes":"", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-c-d/CacheGet.html"},
-		"lucee": {"minimum_version":"", "notes":"For Lucee the method signature is:\n<code>cacheGet( id [, throwWhenNotExist [, cacheName ] ] )</code>\nhowever it does support passing a cacheName in as the second argument for compatibility with ACF.", "docs":"http://docs.lucee.org/reference/functions/cacheget.html"},
+		"lucee": {"minimum_version":"", "notes":"For Lucee the method signature is:\n`cacheGet( id [, throwWhenNotExist [, cacheName ] ] )`\nhowever it does support passing a cacheName in as the second argument for compatibility with ACF.", "docs":"http://docs.lucee.org/reference/functions/cacheget.html"},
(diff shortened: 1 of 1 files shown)


data/en/cacheget.json
```diff
@@ -8,17 +8,16 @@
...
+		{"name":"cacheName","description":"Lucee4.5+ The name of the cache where the object was stored. Applies only to Lucee.","required":false,"default":"","type":"string","values":[]},
+		{"name":"throwWhenNotExist","description":"Lucee4.5+ Enable/Disable throwing an error if element not exists","required":false,"default":"false","type":"boolean","values":["true","false"]}
 	],
 	"engines": {
 		"coldfusion": {"minimum_version":"9", "notes":"", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-c-d/CacheGet.html"},
-		"lucee": {"minimum_version":"", "notes":"For Lucee the method signature is: `cacheGet( id [, throwWhenNotExist [, cacheName ] ] )` however it does support passing a cacheName in as the second argument for compatibility with ACF.", "docs":"http://docs.lucee.org/reference/functions/cacheget.html"},
+		"lucee": {"minimum_version":"", "notes":"For Lucee the method signature is:\n<code>cacheGet( id [, throwWhenNotExist [, cacheName ] ] )</code>\nhowever it does support passing a cacheName in as the second argument for compatibility with ACF.", "docs":"http://docs.lucee.org/reference/functions/cacheget.html"},
```
pfreitag: Should use the backtick instead of HTML -- if the backtick is not converting that description we can fix that.

shaedrich: It does not work as stated in #616.

shaedrich: It's passed to `autoLink()` and `encodeForHTML()` function but nevertheless they stay backticks: https://github.com/foundeo/cfdocs/blob/ee6a1e8b507239987ba7f52726b92a0bfedcf47e/views/doc.cfm#L190-L197

shaedrich: We should run this through [`txtmark.Processor`](/rjeschke/txtmark)

https://github.com/foundeo/cfdocs/blob/9c81f6cd1c986f37fccb57da5ce36b444e419580/doc.cfm#L8-L9

shaedrich: Fixed markdown parsing of `notes`in #622

shaedrich: @pfreitag Removed them.

PR #620, opened 2017-10-19
Cleanup cfinclude

shaedrich:


data/en/cfinclude.json
```diff
@@ -2,34 +2,31 @@
...
-
+		{"name":"template","description":"A logical path to a CFML page.","required":true,"default":"","type":"string","values":[]},
+		{"name":"runonce","description":" CF10+ If set to true, the given template is not processed again for a given request if it has already been processed.","required":false,"default":"false","type":"boolean","values":[]}
 	],
 	"engines": {
-		"coldfusion": {"minimum_version":"", "notes":"CF9: Added the script syntax: include.\nCF10: Added runonce attribute.\nCF11: Changed behaviour such that only files with the extension cfm or cfml are compiled and executed by cfinclude (configurable in Application.cfc via this.compileextforinclude), all other files will be statically included.", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-i/cfinclude.html"},
+		"coldfusion": {"minimum_version":"", "notes":"CF11: Changed behaviour such that only files with the extension cfm or cfml are compiled and executed by cfinclude (configurable in Application.cfc via this.compileextforinclude), all other files will be statically included.", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-i/cfinclude.html"},
```
pfreitag: Why are you deleting `CF9: Added the script syntax: include.\nCF10: Added runonce attribute.\n` ?

data/en/cfinclude.json
```diff
@@ -2,34 +2,31 @@
...
 		"lucee": {"minimum_version":"", "notes":"", "docs":"http://docs.lucee.org/reference/tags/include.html"},
 		"railo": {"minimum_version":"", "notes":"", "docs":"http://railodocs.org/index.cfm/tag/cfinclude"},
 		"openbd": {"minimum_version":"", "notes":"", "docs":"http://openbd.org/manual/?/tag/cfinclude"}
 	},
-	"links": [
-
-	],
+	"links": [],
 	"examples": [
```
pfreitag: These examples should have `"runnable":false` since they will not work on trycf.

data/en/cfinclude.json
shaedrich: It was hard to read so I set the `CF*+` keys at the corresponding places.

data/en/cfinclude.json
shaedrich: @pfreitag Done

data/en/cfinclude.json
```diff
@@ -2,37 +2,35 @@
...
 		},
-	        {
-	            "title": "Tag Syntax",
-	            "description": "",
-	            "code": "<cfinclude template=\"mypage.cfm\">",
-	            "result": ""
-	        }
+		{
+			"title": "Tag Syntax",
+			"description": "CF9+",
```
pfreitag: This is not CF9+, this works in all versions of CFML

data/en/cfinclude.json
pfreitag: Yeah I think the CF9+ was added on the wrong example... I would still like to keep that CF9 added script syntax in the engine notes since I think that is useful, I'm ok with the CF10 one being documented in the attributes.

shaedrich: Yeah I changed the position of CF9+.

For notes the line breaks do not work as discussed in #618.

shaedrich: It seems that the line breaks are not converted properly. They are replaced with a space instead. Should `paragraphFormat()` used instead or additionally or is there anything else wrong?

PR #622, opened 2017-10-19
Markdown parser

shaedrich:
- #616 
- #618 
- #620


pfreitag:
I'm going to decline this PR, right now we support backticks in the autoLink function, I don't want to support full markdown in these areas, I think it will complicate things in the long run.

Thanks for the effort in building this, and all the work you have done thus far!

shaedrich:
So something went's wrong in `autoLink`. In the referenced pr we found out, that backticks do not work in that case. Do you have any idea, why this happens?

shaedrich:
Resolved with 32bf369af7a4b08a9ce8065aa53822a038e6d976.

shaedrich:
Resolved with 32bf369af7a4b08a9ce8065aa53822a038e6d976.

PR #626, opened 2017-10-20
[GUIDE] Application.cfc

shaedrich:


guides/en/applicationcfc.md
```diff
@@ -0,0 +1,87 @@
...
+[2]: https://wikidocs.adobe.com/wiki/display/coldfusionen/onApplicationEnd
+[3]: https://wikidocs.adobe.com/wiki/display/coldfusionen/onSessionStart
+[4]: https://wikidocs.adobe.com/wiki/display/coldfusionen/onSessionEnd
+[5]: https://wikidocs.adobe.com/wiki/display/coldfusionen/onRequestStart
+[6]: https://wikidocs.adobe.com/wiki/display/coldfusionen/onRequest
+[7]: https://wikidocs.adobe.com/wiki/display/coldfusionen/onCFCRequest
+[8]: https://wikidocs.adobe.com/wiki/display/coldfusionen/onRequestEnd
+[9]: https://wikidocs.adobe.com/wiki/display/coldfusionen/onAbort
+[10]: https://wikidocs.adobe.com/wiki/display/coldfusionen/onError
+[11]: https://wikidocs.adobe.com/wiki/display/coldfusionen/onMissingTemplate
```
pfreitag: We _should_ have doc's for all these methods, should link locally. 

Also not sure if this needs to be split into a guide... is it more useful to have the example component on the `Application.cfc` doc instead of in a separate guide where it will be harder to find?

shaedrich: There are more comments than code. This example was pretty huge. That did not look very well.

This methods should be emphasized as lifecycle hooks in my opinion...

pfreitag: I think the idea of a `application` guide might make a lot of sense... It could explain what Application.cfc is, how it is looked for, talk about `Application.cfm` and `OnRequestEnd.cfm` (discouraged), etc. Do you want to rename it as `application.md` and we can cover all things CFML application and link to the various sub docs from there?

shaedrich: That's a good point. There is already a guided for [application scope](https://cfdocs.org/application-scope). Should this one be used for that so both can be merged?

pfreitag: Yeah I think application scope could be merged into the application guide.

shaedrich: Okay, I'll do so :+1:

pfreitag:
Great, thanks!

shaedrich:
Anytime.

PR #640, opened 2017-10-23
extract - Added Description / Values / Examples

Xcreen:
Added Description / Values / Examples


pfreitag:
Thanks - any idea what version of Lucee added this?

Xcreen:
I guess it was also in Railo  (due the header-comment in this file (https://github.com/lucee/Lucee/blob/master/core/src/main/java/lucee/runtime/functions/system/Extract.java)).
So it should be available in all Lucee Version.

PR #649, opened 2017-10-23
Update cachekeyexists.json

Miguel-F:
Added description

### DIFF
--- a/data/en/cachekeyexists.json
+++ b/data/en/cachekeyexists.json
@@ -4,13 +4,13 @@
 	"syntax": "cacheKeyExists(key [, cacheName])",
 	"returns": "boolean",
 	"related": [],
-	"description": "This function is deprecated, use cacheIdExists instead. Returns true/false whether the cache contains an element with the specified keyname.",
+	"description": "Returns true/false whether the cache contains an element with the specified keyname.",
 	"params": [
 		{"name": "key", "description": "Key to check", "required": true, "default": "", "type": "string", "values": []},
 		{"name": "cacheName", "description": "Definition of the cache used by name, when not set the \"default Object Cache\" defined in Lucee Administrator is used instead.", "required": false, "default": "", "type": "string", "values": []}
 	],
 	"engines": {
-		"lucee": {"minimum_version": "", "notes": "", "docs": "http://docs.lucee.org/reference/functions/cachekeyexists.html"}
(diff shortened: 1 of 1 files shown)


data/en/cachekeyexists.json
```diff
@@ -4,14 +4,14 @@
...
-	"description": "",
+	"description": "This function is deprecated, use cacheIdExists instead. Returns true/false whether the cache contains an element with the specified keyname.",
 	"params": [
-		{"name": "key", "description": "", "required": true, "default": "", "type": "string", "values": []},
-		{"name": "cacheName", "description": "", "required": false, "default": "", "type": "string", "values": []}
+		{"name": "key", "description": "Key to check", "required": true, "default": "", "type": "string", "values": []},
+		{"name": "cacheName", "description": "Definition of the cache used by name, when not set the \"default Object Cache\" defined in Lucee Administrator is used instead.", "required": false, "default": "", "type": "string", "values": []}
 	],
 	"engines": {
 		"lucee": {"minimum_version": "", "notes": "", "docs": "http://docs.lucee.org/reference/functions/cachekeyexists.html"}
```
pfreitag: We have a special way to indicate that something is deprecated, you should add `"deprecated":"4.5"` to the `"lucee"` engine object. Assuming it was deprecated in Lucee 4.5

Miguel-F: Sorry about that. I tried searching the Lucee documentation for when it was deprecated but cannot find any of the previous version docs. I would assume it was 4.5 as well. Should I just add that in or take it out completely?

pfreitag: Yeah let's just add in 4.5, if someone finds otherwise it can be updated later. Just by doing that it will automatically add a alert box to the docs showing it is deprecated, for example: https://cfdocs.org/parameterexists

shaedrich:
I don't know if it is neccessary for @pfreitag/cfdocs repo but it would be nice if you could provide a summary of your changes in the commit message.

Miguel-F:
@shaedrich Sebastian - I did add to the commits.  It is short and you may have missed it. "Added description" is what I have done.  And for this particular change, also added the deprecated flag.

shaedrich:
@Miguel-F Yeah, you added it to the second field. That's totally okay, but as you may have noticed, the text of the second field is collapsed on most pages so I just see "Updated file x". That was why I wondered and couldn't have all your changes at a glance ;-)

Issue #569, opened 2017-10-08
Validate against JSON schema


KamasamaK:
We should run all of the JSON data files through the [JSON schema](https://github.com/foundeo/cfdocs/blob/master/data/cfdocs.schema.json) to validate them. Once complete, the JSON schema should be integrated into the Travis CI build process so that all future contributions are guaranteed to be valid.

pfreitag:
FYI - I have added this to the travis-ci build, here is an example: https://travis-ci.org/foundeo/cfdocs/jobs/285772841#L320

It does not fail the tests if the schema is invalid right now, it looks like there are a bunch of minor issues to fix before we start doing that.

KamasamaK:
I had gone through several before submitting the schema, and all of the issues I found were that the `type` was not matching the `enum`s provided. This was mostly due to casing, which should be simple to fix. I can go through them a chunk at a time and fix the issues while trying to avoid conflicts.

pfreitag:
That would be great thanks ... it is probably something you could write a script to do, just check the diff's to make sure your pattern doesn't replace part of an example or something.

fileRead(jsonFile).replace(':"String"', ':"string"', "ALL")

KamasamaK:
I came across an issue with validating tags. I may have to modify the schema based on the answer to this question. For tags, the `returns` key:
1. Should not exist
2. Should be the empty string
3. Should be `void`

Option 2 is the only one that would require a schema change, and it is also how all of the files I've look at seem to be set. I would vote option 1 myself.

shaedrich:
Me to :+1:

pfreitag:
Option 1 probably makes the most sense.

KamasamaK:
I will remove the `returns` from tag docs based on the feedback.

Another issue I've come across is whether the `default` and `values` items in `params` should match the param's `type`. In [the README example](https://github.com/foundeo/cfdocs/blob/master/README.md#json-file-documentation), I noticed that the `type` was "boolean" and the `default` was a string, so I made that schema under the assumption that it would always just be a string and not necessarily match the type. I've noticed docs that use booleans or numbers instead of strings for those fields when the param's type is "boolean" or "numeric", respectively. There may be other cases I haven't come across, but those are the only ones I've seen. 

I'm not aware of a way to have the schema use conditional types, so the options are that the schema type for those fields:

1. Stays as `string`
2. Only contains simple types: `["string", "number", "boolean"]`
3. Has no restriction

[...]

PR #651, opened 2017-10-24
Changed names to be lowercase

KamasamaK:


pfreitag:
Thanks! 

How is the overall schema validation looking? Are we close to being able to fail builds when the schema validation does not pass?

KamasamaK:
I haven't done a comprehensive validation, but these are the issues I've seen that still need to be addressed:

- Param types. This one is the most involved as the issues are not just case-related and the number of them is quite a bit larger than the return types. I will probably address the case issues separately and before tackling the missing/incorrect ones.
- Remove `returns` from tag docs per #569 
- The type for `default` and `values` items also brought up in #569

PR #653, opened 2017-10-24
[CLEANUP] Speed up type tests

### LINKED ISSUE #593: Only test if type matches
Only files with the type used inside one of the tests should be passed in so we just need to loop over them without having to filter any unwanted files to filter with a conditional inside [TestIndexes](/foundeo/cfdocs/tree/master/tests/tests/TestIndexes.cfc).

https://github.com/foundeo/cfdocs/blob/a18feb8dc8a751d3210196812b67cd22dc664856/tests/tests/TestIndexes.cfc#L39-L41

shaedrich:
Resolve #593: Only test if type matches

### DIFF
--- a/tests/tests/TestIndexes.cfc
+++ b/tests/tests/TestIndexes.cfc
@@ -52,7 +52,7 @@ component extends="testbox.system.BaseSpec" {
 			});
 
 			it("should be in functions.json if it is a function", function() {
-				for (function in types.functions) {
+				for (func in types.functions) {
 					expect(ArrayFind(funcs.related, function) NEQ 0).toBeTrue("#json.name# was not in functions.json index");
 				}
 			});
--- a/tests/tests/TestIndexes.cfc
+++ b/tests/tests/TestIndexes.cfc
@@ -29,36 +29,24 @@ component extends="testbox.system.BaseSpec" {
 			tagFiles = arrayFilter(files,function(item) {
 				var json = fileRead(item);
 				return isJSON(json) ? structKeyExists(deserializeJSON(json), "type") && deserializeJSON(json).type is 'tag' : false;
+			}).map(files(function(item) {
+				return deserializeJSON(fileRead(item));
 			});
 			it("should be in tags.json if it is a tag", function() {
-				for (filePath in tagFiles) {
-					var json = fileRead(filePath);
-					var isItJson = isJSON(json);
-					var fileName = getFileFromPath(filePath);
-					if (isItJson) {
-						json = deserializeJSON(json);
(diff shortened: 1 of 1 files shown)


pfreitag:
I don't think this change will speed up the tests... there will be more `fileRead` operations.

shaedrich:
It will prevent the test to loop over files which do not match the type needed.

pfreitag:
Let's assume that there are 500 files, 200 tags, 300 functions. With this code you would do 500 fileReads, then 200 fileReads, then 500 fileReads then 300 fileReads for a total of 1500. The code as it was would only do 1000.

shaedrich:
Well, I removed some duplicated parts.

But did you understand my intention? Maybe you know how to combine it with the origin file to improve it...

pfreitag:
Yeah I understood your intention... maybe something like this:
```
types = {tags=[], functions=[], listings=[]};
for (filePath in files) {
		json = fileRead(filePath);
		isItJson = isJSON(json);
		fileName = getFileFromPath(filePath);
		if (isItJson) {
			json = deserializeJSON(json);
			if (structKeyExists(json, "type")) {
				if (json.type IS "tag") {
					arrayAppend(fileTypes.tags, json.name);
				} else if (json.type IS "function") {
					arrayAppend(fileTypes.functions, json.name);
				} else if (json.type IS "listing") {
					arrayAppend(fileTypes.listings, json.name);
				}
			}
		}
}
```

Then you can just go through the types.tags array, etc. instead of reading the file each time.

shaedrich:
That would be even better. I thought too complicated.

shaedrich:
By the way: Sorry for causing inconvenience and time-consuming discussions...

shaedrich:
(!)JSON Index Tests
       (+)should be in index.json if it is a tag or a function (1605 ms)
       (+)should be in tags.json if it is a tag (31 ms)
       (!)should be in functions.json if it is a function (23 ms)
	      -> Failure: yesNoFormat was not in functions.json index

Now we are getting closer to the point.

shaedrich:
Do we really need to check if it `isJSON` because we already to it in [TestDataFormat](/foundeo/cfdocs/tree/master/tests/tests/TestDataFormat.cfc)?

shaedrich:
@pfreitag Do we really need to check if it `isJSON` because we already to it in [TestDataFormat](/foundeo/cfdocs/tree/master/tests/tests/TestDataFormat.cfc)?

shaedrich:
@pfreitag Are you still with me?

shaedrich:
After all: It works!

shaedrich:
@pfreitag Could you just react to some issues in between?

PR #657, opened 2017-10-24
Move beta state notice to engine notes

shaedrich:


pfreitag:
Hmm what about using the `discouraged` key for that warning?

shaedrich:
Discouraged sounds like it was already in use and should not be used any longer, doesn't it?

KamasamaK:
That is what discouraged is often used for, but the meaning is not restricted to that usage.

Since you're already working on this, you should improve the wording as "experimentell" is not an English word. I would also recommend changing "productive" to "production".

shaedrich:
Yeah, I thought so, too. Just forget that after moving the string.

pfreitag:
The goal of adding `discouraged` notes was to help inform people using a particular feature that there might be a better way of doing what they were trying to do. It is mostly used on the CFUI tags like `cfdiv`, etc. So it is an opinion statement where we can say this should probably be deprecated but it hasn't been.  

I think in this case a discouraged flag still makes sense. We are giving an opinion that you shouldn't use this on production, but its ok to use for debugging or something. Kind of like `System.gc()` I would _discourage_ you from using that in production, but it can be handy for debugging sometimes.

shaedrich:
I understand, what this flag is used for. I just quarrelled with using it in that case.

data/en/sizeof.json
```diff
@@ -4,14 +4,14 @@
...
 	"related": [],
-	"description": "this function is experimentell, to not use in productive enviroment. this function returns the size in bytes of a given object",
+	"description": "Returns the size in bytes of a given object",
 	"params": [
 		{"name": "obj", "description": "", "required": true, "default": "", "type": "any", "values": []},
 		{"name": "complex", "description": "", "required": false, "default": "", "type": "boolean", "values": []}
 	],
 	"engines": {
-		"lucee": {"minimum_version": "", "notes": "", "docs": "http://docs.lucee.org/reference/functions/sizeof.html"}
+		"lucee": {"minimum_version": "", "notes": "", "discouraged":"This function is in an early state. Do not use in production!","docs": "http://docs.lucee.org/reference/functions/sizeof.html"}
```
pfreitag: The `discouraged` key goes on the top level object, not the engine object since it is an opinion covering the feature. Example: https://github.com/foundeo/cfdocs/blob/b6d6968e86f6e105da8430091e39fc798266aa38/data/en/cfpod.json#L8

pfreitag:
My example `cfdiv` does not actually have the discouraged key on it.

pfreitag:
Thanks

shaedrich:
You are welcome

PR #668, opened 2017-10-25
Added basic function description to getfunctionkeywords().

mborn319:
I happened to notice this `getfunctionkeywords()` function is missing a description. (To say the least.) I updated this with the description from the Lucee docs.

http://docs.lucee.org/reference/functions/getfunctionkeywords.html

### DIFF
--- a/data/en/getfunctionkeywords.json
+++ b/data/en/getfunctionkeywords.json
@@ -4,7 +4,7 @@
 	"syntax": "getFunctionKeywords()",
 	"returns": "array",
 	"related": [],
-	"description": "",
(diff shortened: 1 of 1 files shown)


shaedrich:
Is it just me, or is the description from the Lucee docs not that description but rather confusing?

Interesting side note: If I call this function in trycf using Lucee 5, it throws a NullPointerException. With 4.5 all works as expected.

pfreitag:
@mborn319 thanks for updating this... but I'm still not clear sure what this function does, do you think you can explain it further?

pfreitag:
Merged, though I'm still not really sure what the function does.

PR #670, opened 2017-10-25
update cachegetmetadata

bferdinandus:
I added the optional params in the correct order. This is different than in the official coldfusion documentation. To find this I have been digging around in the generated template class files.

### DIFF
--- a/data/en/cachegetmetadata.json
+++ b/data/en/cachegetmetadata.json
@@ -1,7 +1,7 @@
 {
 	"name":"cacheGetMetadata",
 	"type":"function",
-	"syntax":"cacheGetMetadata(id)",
+	"syntax":"cacheGetMetadata(id [, template, region])",
 	"returns":"struct",
 	"related":[],
 	"description":" Gets the metadata values for a cached object.",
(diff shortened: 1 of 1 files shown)


KamasamaK:
Could you please also add those to the `params` array?

shaedrich:
You could omit "(Optional)" in `description` as well because that will already be indicated by setting `required` to `false`.

data/en/cachegetmetadata.json
```diff
@@ -1,13 +1,14 @@
...
+	"syntax":"cacheGetMetadata(id [, template] [, region])",
 	"returns":"struct",
 	"related":[],
 	"description":" Gets the metadata values for a cached object.",
 	"params": [
-		{"name":"id","description":"No Help Available","required":true,"default":"","type":"string","values":[]}
-
+		{"name":"id","description":"The ID of the cached object.","required":true,"default":"","type":"string","values":[]},
+		{"name":"template","description":"Gets metadata for template caches.","required":false,"default":"","type":"string","values":[]},
+		{"name":"region","description":"(Optional) The name of the cache region","required":false,"default":"","type":"string","values":[]}
```
shaedrich: You could omit "(Optional)" in description as well because that will already be indicated by setting required to false.

pfreitag:
Thanks

PR #676, opened 2017-10-26
added descriptions

penhorwood:
Most of these I pulled from the Lucee documentation


data/en/nullvalue.json
```diff
@@ -4,7 +4,7 @@
 	"syntax": "nullValue()",
 	"returns": "any",
 	"related": [],
-	"description": "",
+	"description": "returns null",
```
shaedrich: `returns` is `any` but `description` says it returns `null`?

data/en/nowserver.json
```diff
@@ -4,7 +4,7 @@
 	"syntax": "nowServer()",
 	"returns": "date",
 	"related": [],
-	"description": "",
+	"description": "This function is deprecated. Returns the current time on the server independent of Lucee timezone definition",
```
shaedrich: Move "This function is deprecated." from `description` to `engines.lucee.deprecated`.

shaedrich:
@penhorwood There are some things to fix? Are you still here?

PR #678, opened 2017-10-26
Check related

### LINKED ISSUE #660: [META] Test related
Could we additionally [test](/foundeo/cfdocs/tree/master/tests/tests/TestIndexes.cfc) if it

- relates to itself
- relates to an existing file

If someone links something else than a doc, expect a guide or a listing, what should happen? Is there any option for the user to get that type emphasized?

shaedrich:
Resolve #660

### DIFF
--- a/data/en/cfdirectory.json
+++ b/data/en/cfdirectory.json
@@ -3,7 +3,7 @@
 	"type":"tag",
 	"syntax":"<cfdirectory directory=\".\" action=\"list|create|delete|rename\">",
 	"script": "directoryList(\".\", false, \" \");",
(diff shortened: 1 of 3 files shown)


shaedrich:
@pfreitag Now the question is if listings should be able to relate to non-existing functions/tags. They do not need to be outputted on listing page but could listed in a report for unexisting pages.

Also type "[component](/foundeo/cfdocs/issues/680)" is a special case...

shaedrich:
Not, it works but crashes with application-cfc.

shaedrich:
@pfreitag If #680 is resolved, I can go on depending on the conclusion we came to there.

shaedrich:
@pfreitag It works because it already finds functions and tags related which does not exist.

PR #681, opened 2017-10-26
[CLEANUP] They did all the same

shaedrich:


KamasamaK:
I'm sure what impact this has, but strictly speaking this is not logically equivalent. In the original, when `Len(url.name) LTE 3` is `true` it would not check the `cfelseif` conditions. Now, if `Len(url.name) LTE 3` is `true` and `FindNoCase(url.name, i)` is `false` it will check the other conditions.

shaedrich:
But it would be executed at most once.

shaedrich:
Okay, I guess, the idea was not wrong but I missed your point. So it this pr should just be altered a bit...

pfreitag:
Going to close this one as the `if` statement is not equivalent.

PR #686, opened 2017-10-26
Clearified PI in ColdFusion

shaedrich:


data/en/pi.json
```diff
@@ -4,25 +4,34 @@
 	"syntax":"pi()",
 	"returns":"numeric",
 	"related":[],
-	"description":" Gets the mathematical constant p, accurate to 15 digits.\n The number 3.14159265358979.",
-	"params": [
-	],
+	"description":" Gets the mathematical constant p, accurate to 11 digits.\n The number 3.14159265358979.",
```
pfreitag: Hmm, I think it still returns the 15 digits but when you output it, you only see 11, right?

shaedrich: Yes, but why?

PR #689, opened 2017-10-29
Add description to cfusion_encrypt

caramelomartins:
This pull request adds the description to the function cfusion_encrypt.


pfreitag:
Thanks @caramelomartins - looks good - we should also add the `discouraged` key to this function since it should not be used. Are you able to add that as well?

Here's an example how using discouraged: https://github.com/foundeo/cfdocs/edit/master/data/en/cfdiv.json

caramelomartins:
@pfreitag Yes, sure! I'll do it later today!

pfreitag:
Thanks - I merged this PR you can open a new PR with the discouraged keys.

PR #691, opened 2017-11-01
Add related and example

shaedrich:


shaedrich:
@KamasamaK @pfreitag Everything seems fine...

    -> Error: The ACTUAL argument passed to the isFalse function is not of type boolean.
	  -> Exception Trace: coldfusion.runtime.UDFMethod$InvalidArgumentTypeException: The ACTUAL argument passed to the isFalse function is not of type boolean.

data/en/issimplevalue.json
```diff
@@ -3,20 +3,41 @@
...
 		"railo": {"minimum_version":"", "notes":"", "docs":"http://railodocs.org/index.cfm/function/issimplevalue"},
 		"openbd": {"minimum_version":"", "notes":"", "docs":"http://openbd.org/manual/?/function/issimplevalue"}
 	},
-	"links": [
-
-	]
-
+	"links": [],
+	"examples": [{
+		"title": "",
```
pfreitag: Needs titles on these examples

data/en/settimezone.json
```diff
@@ -3,14 +3,20 @@
...
 	},
-	"examples": [],
-	"links": []
-}
\ No newline at end of file
+	"links": [],
+	"examples": [{
+		"title": "Change Timezone from UTC to CET",
+		"description": "",
+		"code": "writeOutput(getTimezone());\nsetTimezone(\"CET\");\nwriteOutput(\"→ \"&getTimezone());",
```
pfreitag: Hmm this example is not working for some reason on travis:

gettimezone.json example result is:Etc/UTC but evaluated to:EXCEPTION in example 1: Can't cast Object type [sun.util.calendar.ZoneInfo] to a value of type [timezone]. Expected [Etc/UTC] Actual [EXCEPTION in example 1: Can't cast Object type [sun.util.calendar.ZoneInfo] to a value of type [timezone]]

It seams to work on trycf - I wonder if a bug in the specific version of Lucee we are testing with.

shaedrich: It seems that it returns the result of `getTimezoneInfo()` which is a structure instead of `getTimezone()` which is a string...

shaedrich: Looks like a bug. Works here as well for example: http://tpcg.io/dhnnU7

shaedrich: But it works for `getTimezone()`'s example, though...

shaedrich: @pfreitag So what should we do? Nevertheless everything else works well as you can see.

shaedrich:
@pfreitag There is still something wrong (as state in https://github.com/foundeo/cfdocs/pull/691#discussion_r148558555). I wrote a workaround but I can't figure out the error message from what travis returns...

shaedrich:
@pfreitag Do you have any idea, how to fix that?

shaedrich:
In Lucee everything works well.

shaedrich:
@pfreitag Are you still with me. There is the risk that the pr will get more and more outdated.

pfreitag:
Yes still here, just haven't had a chance to look into why this one is failing... Let me try updating testbox to see if that helps.

shaedrich:
No offense! I just wasn't sure.

pfreitag:
Ok I made a change to the test that I think should fix this, let me try merging and see if it is ok.

shaedrich:
Thanks!

PR #692, opened 2017-11-01
Right placement of deprecation notice

shaedrich:
Add description and example


shaedrich:
@pfreitag It seems that auto indexing sometimes fails...

PR #693, opened 2017-11-02
Add discouraged key to cfusion_Decrypt and cfusion_Encrypt

caramelomartins:
Added `discouraged` keys to both `cfusion_Decrypt` and `cfusion_Encrypt`.

@pfreitag is this what you were looking for?


pfreitag:
Thanks - that's good, however I would also discourage the use of `cfmx_compat` as well, I'll update to note that.

PR #696, opened 2017-11-07
[ENHANCEMENT] By Engine

### LINKED ISSUE #583: By Engine
Should we add "functions and tags by engine" to navigation?

Then we would have something like "[Lucee > Unsupported Tags & Functions](http://docs.lucee.org/reference/unsupported.html)" "[Lucee > Additional Tags & Functions](http://docs.lucee.org/reference/additional.html)", but just the supported ones grouped by their engine.

* Adobe ColdFusion
    * an Adobe ColdFusion function/tag
    *  ...
* Lucee
    * an Lucee function/tag
    * ...
* All
    *  an function/tag which is supported by all engines
    *  ...

shaedrich:
Resolve #583


shaedrich:
@pfreitag Could you share your opinion about [this](https://github.com/foundeo/cfdocs/issues/583#issuecomment-363761822) so I can go on?

pfreitag:
Looks good - can you just verify that the code is all working properly and I'll merge it.

shaedrich:
As mentioned in https://github.com/foundeo/cfdocs/issues/583#issuecomment-363761822 I'm unsure which would be the best solution for this.

shaedrich:
@pfreitag It's tested now
image

pfreitag:
Hi @shaedrich this is so old that it now has conflicts, I'm going to close this PR. I think adding so many items to the menu might be too much anyways - what do you think?

shaedrich:
We could just take the first level. That might be helpful.

PR #699, opened 2017-11-08
Update cfthread.json

ianosullivan:
name attribute is required I believe.


shaedrich:
In [BD](http://openbd.org/manual/?/tag/cfthread) and [Lucee](http://docs.lucee.org/reference/tags/thread.html) the `name` attribute is *not* required. in [ACF](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-t/cfthread.html) in the example above, it's seems beeing marked as required but in the table below they say, it's optional...

I tested it in tycf. With Lucee everything works fine without `name` attribute, in ACF there would be an error.

So you should annotate that in engine notes.

pfreitag:
Merged this, created issue #700 to add engine notes.

ianosullivan:
Thanks Seb,

Thanks for that. I should have tested in TRYCF. Didn't think to. Just went
by the engine docs.

Ian.

shaedrich:
No problem, Ian! I made many misstakes here and strained @pfreitag's nerves :D

Seb

pfreitag:
No worries folks - thanks for contributing!

PR #654, opened 2017-10-24
Add issue count to docs

shaedrich:
Resolve #562

PR #703, opened 2017-11-08
Fix docname == undefined

shaedrich:
Fixes #654


pfreitag:
I actually fixed the problem earlier... your change just moves the variable into the expression, but I like that better so I'll merge this anyways. You are probably still seeing undefined because of caching...

shaedrich:
I also added something to the if.

This long-time caching somethimes annoys me a bit if I may mention it...

PR #704, opened 2017-11-10
Since when are getter formatters?

shaedrich:


pfreitag:
Since never... thanks.

PR #705, opened 2017-11-10
Update fileupload.json

KrunchMuffin:
fixed spelling


shaedrich:
Should `accept` have values?

KrunchMuffin:
it could, but the description should be enough.

PR #707, opened 2017-11-10
Update arrayfind.json

bradmarshall:
Updated the description of using arrayFind with a closure / callback function. Previously, it was unclear what arrayFind() would return in the event that an array contains the item being searched for more than once. Updated to indicate that the index returned is the index of the FIRST found item in the array.

### DIFF
--- a/data/en/arrayfind.json
+++ b/data/en/arrayfind.json
@@ -41,7 +41,7 @@
 		},
 		{
 			"title":"Using arrayFind with a closure / callback function",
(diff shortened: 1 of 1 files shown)


shaedrich:
Good job :+1: Then you could reference to reference to `arrayFindAll()` as well.

bradmarshall:
Great idea! Updated to include link to arrayFindAll. It seems like links get auto-parsed which is awesome but I'm new around here so let me know if I did that wrong.

shaedrich:
Everything is well so don't worry :)

Yeah, @pfreitag did a great job :+1:

bradmarshall:
Awesome! Agreed!

pfreitag:
Thanks @bradmarshall 

Congrats on your first contribution to cfdocs!

PR #711, opened 2017-11-13
Add guides to docs

### LINKED ISSUE #614: Add guides to listings and docs
It would be nice if we could internally link guides to `listings` and `docs` with an additional key.

So we could store larger code snippets which involve multiple functions from the same "function family" in the guide covering an overview and simply link it.

Related to #629, #623

shaedrich:
Resolve #614


shaedrich:
This is tested as well.

@pfreitag Do you like that the related guides are positioned right under the doc title while the other related docs are positioned under the breadcrumbs?
image
image

shaedrich:
@pfreitag Did you see this?

PR #713, opened 2017-11-13
Fix error message for test

shaedrich:


pfreitag:
good catch thanks

shaedrich:
I wondered when I got the error that the title existed and so I did a bit of investigation.

PR #726, opened 2017-11-20
 Refactoring docs view

shaedrich:


views/doc.cfm
```diff
@@ -4,21 +4,25 @@
...
+						</cfif>
+					</code>
+				</p>
 				<cfif data.type IS "tag">
 					<cfif StructKeyExists(data, "script")>
 						<cfset data.scriptTitle = "Script Syntax">
 					<cfelseif ListFindNoCase("cfabort,cfbreak,cfcontinue,cfreturn,cfexit", data.name)>
 						<cfset data.scriptTitle = "Script Syntax">
-						<cfset data.script = ReplaceNoCase(data.name, "cf", "") & ";">
+						<cfset data.script = replaceScript(name = data.name, mode = "cf") & ";">
```
pfreitag: This is incorrect here... For tags like `cfabort` the script eqivilent is simply `abort;` so it should remove the "cf" from the tag name. Given that I don't see any need to add the replaceScript function since it is only used once.

shaedrich: I wanted to remove the bunch of cfset tags so you could focus on the major code.

pfreitag: Ah I see you are passing a different mode.

PR #727, opened 2017-11-20
Update truefalseformat.json

creepshow:


data/en/truefalseformat.json
```diff
@@ -4,7 +4,7 @@
 	"syntax": "trueFalseFormat(value)",
 	"returns": "string",
 	"related": ["yesNoFormat", "isBoolean","booleanFormat"],
-	"description": "",
+	"description": "Returns a boolean value of true or false for the argument passed in.",
```
shaedrich: I would add "as string" after "boolean value" because `returns` indicates it as `string`.

shaedrich: Already done in https://github.com/foundeo/cfdocs/pull/676/commits/110d71c95506483225b893313bf076e76fd3ef84#diff-0f967c1ed3408de69e2971c887705333R7?

Or is that description better?

PR #731, opened 2017-11-23
Add CFML Form Tags listing

shaedrich:


pfreitag:
Do you think we need a category for this? I would think these tags would fit well under `ui-tags`

shaedrich:
Okay then I will move it there.

PR #733, opened 2017-11-24
Added compatibilty note for script version of tag

bradmarshall:
Calling cfheader(), the script-based version of this tag, is only available on CF11+. CF10 will trigger a 500 "Variable CFHEADER is undefined" error.

### DIFF
--- a/data/en/cfheader.json
+++ b/data/en/cfheader.json
@@ -13,7 +13,7 @@
 
 	],
 	"engines": {
-		"coldfusion": {"minimum_version":"", "notes":"", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-g-h/cfheader.html"},
+		"coldfusion": {"minimum_version":"", "notes":"Script-based cfheader() added in CF11", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-g-h/cfheader.html"},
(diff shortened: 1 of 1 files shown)


data/en/cfheader.json
```diff
@@ -13,7 +13,7 @@
 
 	],
 	"engines": {
-		"coldfusion": {"minimum_version":"", "notes":"", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-g-h/cfheader.html"},
+		"coldfusion": {"minimum_version":"", "notes":"Script-based cfheader() added in CF11", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-g-h/cfheader.html"},
```
shaedrich: You could add a plus sign, so it could be transformed into a label properly: `CF11+` and put it at the beginning of notes.

bradmarshall:
Ah, I didn't realize you could transform those version labels like that. Done! Moved label to front of note as well.

shaedrich:
Yeah, that's pretty nice. So you can have it all at a glance.

KamasamaK:
CFDocs already assumes script syntax is available for tags in ACF11+, Lucee, and Railo 4.2+. If you look at the [CFDocs page for cfheader](https://cfdocs.org/cfheader), you can see that it displays the syntax near the top next to the lightning bolt icon, and if you hover on that icon it will indicate the engines it became available.

shaedrich:
Oh, that didn't even I know.

PR #736, opened 2017-11-29
Update datediff.json

aliaspooryorik:
Both Lucee and ACF use "w" and "ww" for weeks. There is no weekday option.

As per ACF docs:

String that specifies the units in which to count; for example yyyy requests a date difference in whole years.

yyyy: Years
q: Quarters
m: Months
y: Days of year (same as d)
d: Days
w: Weeks
ww: Weeks
h: Hours
n: Minutes
s: Seconds

cfdocs currently has:

w: Weekday

### DIFF
--- a/data/en/datediff.json
+++ b/data/en/datediff.json
@@ -7,7 +7,7 @@
 	"related":["dateadd","dateformat"],
 	"description":"Determines the integer number of datepart units by which date1 is less than date2.",
 	"params": [
-		{"name":"datepart","description":"yyyy: Year\n q: Quarter\n m: Month\n y: Day of year\n d: Day\n w: Weekday\n ww: Week\n h: Hour\n n: Minute\n s: Second\n","required":true,"default":"","type":"string","values":["yyyy","q","m","y","d","w","ww","h","n","s"]},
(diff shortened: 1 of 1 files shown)


shaedrich:
In BD there is

`y`: day of year
`d`: day

and

`ww`: week of year
`w`: week

Whereas Lucee and ACF just use this for `y` and `d` but I guess, I't the same so we should make it consistent.

But because both `y` and `d` & `ww` and `w` are the same we could simply say:

`y`/`d`: day
`ww`/`w`: week

shaedrich:
@pfreitag Which notation do you prefer? First or second one or should we leave it as it is now after merging @aliaspooryorik's commit?

pfreitag:
I don't have a strong opinion about it, but I think having each mask on its own line is perhaps easier to read.

shaedrich:
Then `ww` should be "week of year" to be consistent, shouldn't it?

pfreitag:
The _week of year_ only applies to OpenBD right? I think OpenBD usage is quite low so that might make better sense as an engine note.

shaedrich:
BD says as well as Lucee and ACF that it is the same as without "of year" just like with "day of week".

Issue #363, opened 2016-06-10
Have `CF+`, or `Lucee+` work to indicate an argument or feature that is only supported by a specific engine.


mjhagen:
Would work the same way as `CF9+` just without the version number.

pfreitag:
That would be handy in some cases I think so I'm not opposed to adding that. I do prefer if the version number is included if possible since that is useful information in most cases.

mjhagen:
I ran into one on directoryList, the `type` argument is only available on ACF, I wondered if I should add a note for Lucee/Railo/BD, or one for ACF and thought that just having a label on the argument itself would be clearer (right now I added CF9+ on that arg.)

timbeadle:
👍  on this idea. Similar to the browser table on e.g. Mozilla Developer Network: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

timbeadle:
(& I was tripped up by string.replaceList() not being implemented on Lucee 4.5)

shaedrich:
Keep in mind that features can be removed (e. g. CF11-). Should it be releated to engines.minimum_version?

KamasamaK:
An `engines` object would be good for this, but it would need to relate to each parameter instead of the entity as a whole. Also, @shaedrich brings up a good point. The `engine` object should also indicate when something was deprecated or removed, in addition to when it was introduced. Perhaps something like
```
"params": [
	{
		"name":"funcArgNameOrTagAttributeName",
		"description":"What it does",
		"required":true,
		"default":"false",
		"type":"boolean",
		"values":["true","false"],
		"engines": {
			"coldfusion": {"minimum_version":"10", "deprecated_version":"", "removed_version":"", "notes":"CF Specific Info Here"},
			"railo": {"minimum_version":"4.1", "deprecated_version":"", "removed_version":"", "notes":"Railo Specific Here"},
			"lucee": {"minimum_version":"4.5", "deprecated_version":"", "removed_version":"", "notes":"Lucee Specific Info Here"}
		}
	}
]
```

shaedrich:
`-` could indicate a removed feature, but how about a _deprecated_ one? (@KamasamaK good idea to add deprecations as well)

shaedrich:
This should be done for values as well.

pfreitag:
[...]

PR #738, opened 2017-12-01
 mention cf_sql_int oddities

coryasilva:


KamasamaK:
`engines` is actually added within the "cfsqltype" param and not between params as stated. I would figure that at worst it would just be ignored and not break anything, but I'm not familiar with how the JSON is parsed for the web application. I did in fact have `engines` able to be used there in the [cfdocs.schema.json](https://github.com/foundeo/cfdocs/blob/master/data/cfdocs.schema.json) based on [a comment](https://github.com/foundeo/cfdocs/issues/363#issuecomment-327831608) by @pfreitag thinking it would eventually be supported.

pfreitag:
Correct - I don't think it will actually _break_ anything to have an `engines` key in the `params`  there, but it is not officially supported yet (not used anywhere).  @KamasamaK is correct - the eventual goal is to support adding engine metadata to params.

shaedrich:
As long as `param.engine` is not officially supported yet, these engine notes should also be added to _global_ `engine.notes`.

PR #748, opened 2017-12-06
Updated description and added example

gratzc:


shaedrich:
Just a small tip: You could group commits with similar changes into one pull request so they could be reviewed all at once.

gratzc:
ha, but then I can't meet my advent 24 pull requests, https://24pullrequests.com/  I choose cfdocs to do my 24 pull requests with.  :)  But if it is seriously a pain for you I can group them, the point is to be helpful

shaedrich:
It's down to you to decide. I'm not the maintainer of this project so you do not have to be considerate of me :D 

Now I know the reason behind it :+1: In my opinion it should not result in spamming like making more pull requests as necessary (not more as you would and/or could normally make) but that's just a personal opinion and negligible^^

gratzc:
I agree, I am just doing one a day and making it improving one of the tag/function docs, so it seemed a logical breakpoint.  Not trying to do one for a word change and another for another small change to just spam.

pfreitag:
@gratzc no worries - I'll take PRs any way they come in.  Thanks for choosing cfdocs to do this!

PR #750, opened 2017-12-06
Correct type

shaedrich:
Fix #676 (https://github.com/foundeo/cfdocs/pull/676#pullrequestreview-72096729)

### DIFF
--- a/data/en/nullvalue.json
+++ b/data/en/nullvalue.json
@@ -2,7 +2,7 @@
 	"name": "nullValue",
 	"type": "function",
 	"syntax": "nullValue()",
-	"returns": "any",
+	"returns": "null",
 	"related": [],
 	"description": "returns null",
 	"params": [],


pfreitag:
Hmm I think using null here would fail the JSON schema validation because `null` is not a CFML argument/return type, so that is why it is "any". 

Care to comment @KamasamaK ?

shaedrich:
Hmmm... Lucee seems to recognize that it's null. So maybe we should add it...

KamasamaK:
You are correct that it would fail the JSON schema validation since it is not one of the enumerated return types. The [Lucee documentation](http://docs.lucee.org/reference/functions/nullvalue.html) also has it as returning `any`. I might even say that `void` could also work in this case, but I'm not sure if Lucee handles nulls in that way and that might not really be clear for the purpose of documentation.

shaedrich:
They would not have a null if it would be completely useless, would they?

pfreitag:
So... `null` is a value, not a type. An object reference of any type can be null in java. In the case of this function you are calling it because you want to pass the `null` value into a java method or return null from a function. So based on that I don't think it makes sense to call `null` it a _return type_ - do you agree?

shaedrich:
I agree in respect of CF data types. But when you dump it, it's `Empty: null`. And any is just unspecified. So it `null` is a lack in CF, I know...

pfreitag:
I don't understand how you can dump `any` it is just a type -- do you mean a function call that is `void`?

shaedrich:
I meant especially with `nullValue()`: https://trycf.com/editor/gist/6990f6b01db616b2eae5b24818076470/lucee5

pfreitag:
So I just did a little test to see what a void function would dump as, and it is the same: 

https://trycf.com/editor/gist/a956964a87c61371258df6af3b6c62d0/lucee5

Also tested on ACF 2016 and `isNull(voidFunc()) == true`

shaedrich:
Hmmm... Okay... A bit inconsistent to call `any` null...

Touché!

pfreitag:
Going to close this PR, I think `any` is more appropriate for the type here.

PR #754, opened 2017-12-08
Updated ormClearSession docs and added example

gratzc:


pfreitag:
Nice work Curt - I'm sure the rest of the ORM functions could use some love as well if you are looking for more.

shaedrich:
If you have finished with them you might like to write a guide regarding ORM. We just have one for the settings: https://cfdocs.org/ormsettings

PR #755, opened 2017-12-08
Added example to fileExists

gratzc:


shaedrich:
Unfortunately we cannot test it. Maybe just create a temporary file?

PR #776, opened 2018-01-25
Cfldap

bferdinandus:
correct typo in parameter description


pfreitag:
That's an important distinction - thanks @bferdinandus!

PR #778, opened 2018-01-29
Adding Example to sum query column

ianosullivan:
Get the total of a query column by using valueArray() and then using the sum() function


ianosullivan:
Not familiar (haven't used) with `queryReduce()`. Thanks for the pointer. I find `valueArray(qryProducts, 'price').sum()` a bit more concise/shorter though than `queryReduce()` but I guess `queryReduce()` is better suited. 
Should have used `querynew()`. I haven't used it ever so didn't think of it.

shaedrich:
No problem. `valueArray(qryProducts, 'price').sum()` looks like a nice approach as well but yes, `reduce` function for queries, arrays, etc. is more standard.

PR #781, opened 2018-02-01
Better example

shaedrich:


pfreitag:
Thanks! BTW the `<cfscript>` tags are not necessary to include in the example it will add them automatically.

shaedrich:
Oh, you're right.

PR #784, opened 2018-02-05
Add related

shaedrich:


data/en/cferror.json
```diff
@@ -2,23 +2,19 @@
 	"name":"cferror",
 	"type":"tag",
 	"syntax":"<cferror type=\"exception\" template=\"\">",
-	"related":[],
+	"related":["cfabort","cfthrow"],
```
pfreitag: `cftry` `cfcatch` and `onError` might be good additions here too.

PR #803, opened 2018-03-01
Add examples

shaedrich:


shaedrich:
@pfreitag 
> Failure: xmlvalidate.json example result is:true but evaluated to:EXCEPTION in example 1 in xmlvalidate.json: Remote host closed connection during handshake

pfreitag:
I restarted the Travis job and it passed this time -- the failure was temporary because it couldn't connect to the URL that it needed to to run the test for xmlValidate.

shaedrich:
Okay, fine.

Issue #751, opened 2017-12-06
cfpop Missing 'secure' argument


mar-hn:
For more information see 
https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-p-q/cfpop.html

donovanclarke:
This issue can be closed :)

PR #805, opened 2018-03-03
#751 Add secure argument

donovanclarke:
Add in secure argument for issue -- #751

### DIFF
--- a/data/en/cfpop.json
+++ b/data/en/cfpop.json
@@ -18,6 +18,7 @@
 		{"name":"maxrows","description":"Number of messages to return or delete, starting with the\n number in startRow. Ignored if messageNumber or uid is\n specified.","required":false,"default":"","type":"numeric","values":[]},
 		{"name":"startrow","description":"First row number to get or delete. Ignored if messageNumber\n or uid is specified.","required":false,"default":"","type":"numeric","values":[]},
 		{"name":"generateUniqueFileNames","description":"Yes: Generate unique filenames for files attached to an\n e-mail message, to avoid naming conflicts when files are\n saved","required":false,"default":"","type":"boolean","values":[true,false]}
+		{"name":"secure","description":"Yes: enables SSL for pop requests.","required":false,"default":"no","type":"boolean","values":[true,false]}
 	],
 	"engines": {
 		"coldfusion": {"minimum_version":"", "notes":"", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-p-q/cfpop.html"},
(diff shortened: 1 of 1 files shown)


shaedrich:
Looks good. Just a few things which could be improved.

data/en/cfpop.json
```diff
@@ -17,7 +17,8 @@
 		{"name":"timeout","description":"Maximum time, in seconds, to wait for mail processing","required":false,"default":"","type":"numeric","values":[]},
 		{"name":"maxrows","description":"Number of messages to return or delete, starting with the\n number in startRow. Ignored if messageNumber or uid is\n specified.","required":false,"default":"","type":"numeric","values":[]},
 		{"name":"startrow","description":"First row number to get or delete. Ignored if messageNumber\n or uid is specified.","required":false,"default":"","type":"numeric","values":[]},
-		{"name":"generateUniqueFileNames","description":"Yes: Generate unique filenames for files attached to an\n e-mail message, to avoid naming conflicts when files are\n saved","required":false,"default":"","type":"boolean","values":[true,false]}
+		{"name":"generateUniqueFileNames","description":"Yes: Generate unique filenames for files attached to an\n e-mail message, to avoid naming conflicts when files are\n saved","required":false,"default":"","type":"boolean","values":[true,false]},
+		{"name":"secure","description":"Yes: enables SSL for pop requests.","required":false,"default":"no","type":"boolean","values":[true,false]}
```
shaedrich: Looks good. Just a few things which could be improved.

- `boolean`s don't explicitly need values `true` and `false`
- If only `YES` is explained in `description` you could also write "Enable/Disable" of just omit the "Yes:" at the beginning.
- `true` would be better than `YES` for consistency. According to this `default` could be `false` rather than `no`.

donovanclarke: I agree that for consistency it seems that using `false` or `true` is better.

 I checked other boolean arguments, and it seems that even though they don't necessarily have to take `true` or `false`. It would go against project consistency to take it out?

shaedrich: Hmmm... In my opinion the possible values are obvious when using parameter type "boolean" so they can be omitted because they give us no additional information.

pfreitag:
Thanks @donovanclarke 

Also it would be good to put `CF10+` in the description of the secure attribute. Also it looks like we are missing this _ ColdFusion 11: New action, markRead is supported. New attribute delimeter _

@donovanclarke if you are up for a few more changes fell free to add the above.

data/en/cfpop.json
pfreitag: Oops - sorry I missed this comment!

Yes when the type is boolean, it doesn't show the values because it is implied that they are in the subset of the possible boolean values.

PR #809, opened 2018-03-05
#751 add cf11 attributes to cfpop

donovanclarke:
@pfreitag : It seems that I did miss some new attributes! I have made an update on the same branch. 

Thanks!


pfreitag:
Thanks @donovanclarke !!

PR #817, opened 2018-03-09
Added details/examples to queryDeleteRow 

danacoder:


shaedrich:
I guess, the "the" at the beginning of the param description can be omitted.
In my opinion "1" instead of "(1)" suffice.
I'm not sure if "modifying the original query" is needed in the description although it should be mentioned somewhere. I guess that's sort of like a generic info...

shaedrich:
Ah, okay. But I think, if the font is the problem, it should solved by changing the font.

Looks fine now.

pfreitag:
Good work - thanks!

PR #826, opened 2018-03-14
Values with and without  `cf_sql_` prefix must be equivalent

shaedrich:


pfreitag:
Hmm -- I know they are equivalent but I'd like to keep those in there for the sake of IDE's that use it for autocompletion.

shaedrich:
So seen from this angle we should add all values present with `cf_sql_` prefix now additionally without for consistency, shouldn't we?

pfreitag:
Yeah probably should.

shaedrich:
Okay, I'll change that.

pfreitag:
@shaedrich FYI CF2016 Failure: cfqueryparam.json was NOT valid JSON

shaedrich:
Done.

PR #831, opened 2018-03-15
Remove optional parameters from tag syntax

shaedrich:
The whole signature broke the layout

### DIFF
--- a/data/en/cfcookie.json
+++ b/data/en/cfcookie.json
@@ -1,7 +1,7 @@
 {
 	"name":"cfcookie",
 	"type":"tag",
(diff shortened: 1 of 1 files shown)


pfreitag:
@shaedrich Maybe we should keep `value` in there because while not required all the time it is used most of the time. What do you think?

shaedrich:
I see your point, but that would break ranks, won't it?

pfreitag:
I think the tag syntax should mainly be useful, I don't think I've had a strict rule to apply as to what is included there. Is it usually just the required attribute that are shown?

shaedrich:
What's a cookie without a `value`?

Maybe we could make a second signature underneath the first one?

pfreitag:
I think the only reason that you can or would omit `value` is if you also have `expires="now"`

shaedrich:
But then set an infinite cookie without a value? So just checking it for existence?

pfreitag:
Merging this, looks like you can also set a cookie without a value, not sure why you would want to do this but its possible.

PR #846, opened 2018-03-30
Structreduce lucee

jeremyhalliwell:


pfreitag:
Thanks @jeremyhalliwell

PR #850, opened 2018-04-10
Add Lucee 4.5 example of array of structs query

gpickin:
Added example of how to return an array of structs in Lucee 4.5
Also updated the existing array of structs to be clear that its for CF11+ and Lucee5+


pfreitag:
Thanks Gavin!

gpickin:
So, after finding this out, and testing, i found an issue.

returnType = array is not valid as mentioned
returnType = array-of-entity is valid but still returns a query
returnType = array-of-entities is valid but still returns a query

So might want to change or remove this.
Sorry.
Thoughts?

pfreitag:
@gpickin Yeah looks like `array-of-entities` does return an array for Lucee 5, but still returns a query for Lucee 4.5 (but does not throw an exception). Not sure why you would use that instead of just `returntype="array"` so it probably makes sense to remove the example _Return Query as an Array of Structs - Lucee 4.5_ agree?

gpickin:
If you have returntype=”array” in 4.5 it will error… which is weird, and why I wanted to note it.

PR #852, opened 2018-04-11
Fix errant required param and param order

ddspringle:
Fixed timeZone param being marked as required and fixed ordering of
params to match expected ordering of the function


pfreitag:
Thanks @ddspringle

PR #854, opened 2018-04-27
Fix typo: speficied  -> specified

laurencehoess:


pfreitag:
Thanks @laurencehoess!

Issue #760, opened 2017-12-19
getLocaleCountry


KamasamaK:
I have been unable to find anything on the function [`getLocaleCountry`](https://cfdocs.org/getlocalecountry) added in #581. All of the `engine` links are for `getLocaleDisplayName` and the example does not work on TryCF for any of the engines.

pfreitag:
@shaedrich - looks like you added those, do you have any more info?

shaedrich:
You can find information about it in the [Lucee update docs](http://docs.lucee.org/guides/lucee-5/functions-tags.html#localeregional-functions). When I added them, they seemed to work as far as I remember.

KamasamaK:
@shaedrich Can you please verify which version of Lucee that you tried that supports these functions? None of the engines hosted on TryCF support these functions.

shaedrich:
As stated in the update docs, they were added and removed in the same version only for internal proposes. So they do not work in any version. Don't ask me for reasons, ask them :D

KamasamaK:
I didn't see anything in the update docs stating they were added and removed in the same version. They were included in a list with no further information on them other than that they were deprecated (which is not the same as being removed). I thought you said you had gotten them to work. That doesn't really make much sense if they were added and removed in the same version, but if so then my previous question regarding which version exactly still needs an answer. 

Otherwise, is it even worth it to keep them documented? The current documentation seems useless as it is -- no useful external references, no way to test, inaccurate `engines` information.

shaedrich:
Oh, sorry. You're right. They are deprecated. That's all I know. We both read the document. That's all we have.
The version is written in the update docs and on cfdocs as well: Lucee 5. Why is that engine information inaccurate. Sorry but I don't see your point.

The question of keeping it documented may be addressed to @pfreitag. But for me it doesn't really matter matter if we keep it. I just tried to get rid of the open issue where the update docs were linked.

KamasamaK:
@shaedrich The `engines` is inaccurate because it says that all 4 of the engines support it and have no indication of when it was added or removed from any engine, when we now know that if it was ever added to any that it must have been removed since it does not work in any recent versions of those engines. It is also inaccurate because the `docs` links are for a different function that make no mention of the relevant function.

[...]

PR #860, opened 2018-05-16
Fix up wrong information

shaedrich:
- Neither Open BD nor ACF ever supported that function
- Update docs state that function was added and deprecated in that version but gives not much more information

For discussion see: #760

### DIFF
--- a/data/en/getlocalelanguage.json
+++ b/data/en/getlocalelanguage.json
@@ -7,7 +7,7 @@
 	"description":"Gets the language from where the locale belongs to.",
 	"params": [],
 	"engines": {
-		"lucee": {"minimum_version":"", "added": "5.0", "deprecated":"5.0", "notes":"Lucee5+ Deprecated in favor of getLocaleInfo()", "docs":"http://docs.lucee.org/reference/functions/getlocaledisplayname.html"},
+		"lucee": {"minimum_version":"", "added": "5.0", "deprecated":"5.0", "notes":"Lucee5+ Deprecated in favor of getLocaleInfo()", "docs":"http://docs.lucee.org/reference/functions/getlocaledisplayname.html"}
 	},
 	"links": [],
 	"examples": [{
--- a/data/en/getlocalelanguage.json
+++ b/data/en/getlocalelanguage.json
@@ -7,10 +7,7 @@
 	"description":"Gets the language from where the locale belongs to.",
 	"params": [],
 	"engines": {
-		"coldfusion": {"minimum_version":"", "notes":"", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-e-g/getlocaledisplayname.html"},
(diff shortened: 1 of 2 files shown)


KamasamaK:
The same issue also applies to `getLocaleLanguage`.

shaedrich:
You're right. Fixed it.

data/en/getlocalelanguage.json
```diff
@@ -7,10 +7,7 @@
 	"description":"Gets the language from where the locale belongs to.",
 	"params": [],
 	"engines": {
-		"coldfusion": {"minimum_version":"", "notes":"", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-e-g/getlocaledisplayname.html"},
-		"lucee": {"minimum_version":"", "deprecated":"5", "notes":"Lucee5+ Deprecated in favor of getLocaleInfo()", "docs":"http://docs.lucee.org/reference/functions/getlocaledisplayname.html"},
-		"railo": {"minimum_version":"", "notes":"", "docs":"http://railodocs.org/index.cfm/function/getlocaledisplayname"},
-		"openbd": {"minimum_version":"", "notes":"", "docs":"http://openbd.org/manual/?/function/getlocaledisplayname"}
+		"lucee": {"minimum_version":"", "added": "5.0", "deprecated":"5.0", "notes":"Lucee5+ Deprecated in favor of getLocaleInfo()", "docs":"http://docs.lucee.org/reference/functions/getlocaledisplayname.html"},
```
pfreitag: Need to remove the trailing comma here for it be valid JSON

KamasamaK:
The `docs` links are still pointing to a different function. Since the relevant docs don't actually exist, if they _have_ to point to a function, I think it makes more sense to use `getLocaleInfo`, but I believe they are not required. Also, they don't have a `removed` version, which is the main issue since it is not usable in the latest version.

data/en/getlocalelanguage.json
shaedrich: Oh, you're right.

shaedrich:
In terms of "using" `getLocaleInfo` is of course the right and only choice. Noone ever said, you should use `getLocaleLanguage`.

@pfreitag To make a long story short, do we need it or do we not?

KamasamaK:
@shaedrich You misunderstand. I was referring to what to use for the `docs` links, not what a developer uses.

shaedrich:
@KamasamaK Oh, okay, sorry, I actually did. So link to from `getLocaleLanguage`'s cfdocs page to `getLocaleInfo`'s Lucee page?

KamasamaK:
Yes, linking to `getLocaleInfo` was my suggestion over using `getLocaleDisplayName` which is what both are currently using. However, I still think it preferable to not include a `docs` link if no documentation for that function exists.

shaedrich:
So do we remove the link change it to `getLocaleInfo`.

pfreitag:
Sorry for delay in responding ... so if you remove the docs link it will not show the Lucee 5+ indicator on the top right... This may be ok however since it is deprecated right?

shaedrich:
Well, makes sense. I do so.

pfreitag:
Thx

Issue #824, opened 2018-03-13
directorySize


CoherentLogic:
This function does not actually exist in Lucee or ACF.

KamasamaK:
@pfreitag This can be closed now

pfreitag:
Fixed in #865

PR #865, opened 2018-06-07
Remove Function DirectorySize

Xcreen:
Remove Function DirectorySize  
Issue: #824 
Could not find any docs: 
https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-c-d.html
http://openbd.org/manual/?/function/
http://docs.lucee.org/reference/functions.html

Also dont work with trycf (function dont exist)


pfreitag:
Hmm interesting, looks like it should be removed.

Any chance you could rebuild the index files again, they are now conflicting due to another PR I just merged.

pfreitag:
Thanks!

Issue #867, opened 2018-06-08
queryExecute Return Query as an Array of Structs not supported in CF11+


rejithtechver:
In the examples it is mentioned that **Return Query as an Array of Structs - Lucee5+ CF11+** 
![image](https://user-images.githubusercontent.com/4943104/41157171-60bd0900-6b42-11e8-86a9-b44f9a26ff2f.png)

This is not supported in Adobe CF. You can see this [here](https://www.trycf.com/gist/6977fbbf91906b9b9f1aa22a013ed61f/acf2018?theme=monokai).

shaedrich:
For me it all works, expect CF 10 (`queryExecute` wasn't supported yet), Railo and Lucee 4.5 (attribute `returntype` of tag query has an invalid value).

PS: From which year (ABY) did you take their last names? :D

KamasamaK:
It may run fine, but it doesn't produce the expected output. In CF11+, `returntype` is ignored unlike Railo and Lucee 4.5 where it produces an error.

shaedrich:
Okay, then we have to put a not in that doc?

rejithtechver:
Not really, we just need to get rid of "CF11+" from the docs. Check the screenshot attached to the OP.

shaedrich:
Not the whole. `queryExecute` still works. Only one parameter does not.

rejithtechver:
Yes, what I am trying to tell you is that the example "Return Query as an Array of Structs - Lucee5+ CF11+" gives out wrong information. There isn't anything else wrong with it.

![image](https://user-images.githubusercontent.com/4943104/41273565-9bc77d8a-6e37-11e8-9257-cf1976c105ee.png)

pfreitag:
@rejithtechver Thanks for reporting this -- this is a pretty straight forward change to make if you would like to try to do it yourself, we'll guide you through the process.

1) Find the json file for queryExecute in the `data/en` directory: https://github.com/foundeo/cfdocs/blob/master/data/en/queryexecute.json
2) Click on the Pencil icon to edit the file
3) Make the change and click the button to send the pull request.
4) Your pull request will show up here: https://github.com/foundeo/cfdocs/pulls where I can merge it in if everything looks good.

If you have any questions feel free to reach out - and let me know if you would rather not make the change yourself so someone else can go ahead and do it.

rejithtechver:
Oh, sorry Pete, I didn't know I could suggest a change from my end. Thank you for the info.

[...]

PR #868, opened 2018-06-12
Return as Array of Structs not supported in CF11+

rejithtechver:
https://github.com/foundeo/cfdocs/issues/867 Fix for this.


pfreitag:
@rejithtechver perfect - thanks!

PR #869, opened 2018-06-22
Added example to structFind() and optional 'defaultValue' parameter

SebastianZ:
The `defaultValue` parameter is currently only supported by Lucee and their docs don't specify what's the default value for that parameter.

Not sure whether the note for its support should be at the parameter or at the engines. I've mentioned it in the description for the parameter itself, so it's obvious when looking at it.

### DIFF
--- a/data/en/structfind.json
+++ b/data/en/structfind.json
@@ -1,13 +1,14 @@
 {
 	"name":"structFind",
 	"type":"function",
-	"syntax":"structFind(structure, key)",
+	"syntax":"structFind(structure, key [, defaultValue ])",
 	"returns":"any",
-	"related":[],
-	"description":" Determines the value associated with a key in a structure.",
+	"related":["structFindKey", "structFindValue"],
+	"description":"Determines the value associated with a key in a structure.",
 	"params": [
-		{"name":"structure","description":"Structure that contains the value to return","required":true,"default":"","type":"struct","values":[]},
-		{"name":"key","description":"Key whose value to return","required":true,"default":"","type":"string","values":[]}
+		{"name": "structure", "description": "Structure that contains the value to return", "required": true, "default": "", "type": "struct", "values": []},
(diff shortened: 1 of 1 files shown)


data/en/structfind.json
```diff
@@ -1,13 +1,14 @@
...
-	"related":[],
-	"description":" Determines the value associated with a key in a structure.",
+	"related":["structFindKey", "structFindValue"],
+	"description":"Determines the value associated with a key in a structure.",
 	"params": [
-		{"name":"structure","description":"Structure that contains the value to return","required":true,"default":"","type":"struct","values":[]},
-		{"name":"key","description":"Key whose value to return","required":true,"default":"","type":"string","values":[]}
+		{"name": "structure", "description": "Structure that contains the value to return", "required": true, "default": "", "type": "struct", "values": []},
+		{"name": "key", "description": "Key whose value to return", "required": true, "default": "", "type": "string", "values": []},
+		{"name": "defaultValue", "description": "Default value which will be returned if the key does not exist or if null was found. Currently only supported by Lucee. See http://docs.lucee.org/reference/functions/structfind.html#argument-defaultValue", "required": false, "default": "", "type": "any"}
```
shaedrich: You can put Lucee4.5+ engine label at the beginning of the description.

data/en/structfind.json
```diff
@@ -1,13 +1,14 @@
...
 
 	],
 	"engines": {
```
shaedrich: And yes, you were right: Mentioning this in the engine notes for Lucee is a good idea :+1:

data/en/structfind.json
SebastianZ: Done in https://github.com/foundeo/cfdocs/pull/869/commits/d5680f8480cc532f7f8953475d0046f40b5db0e5.

SebastianZ:
I have also added the member function syntax now.

PR #871, opened 2018-06-26
Added descriptions to listEvery() function

SebastianZ:


data/en/listevery.json
```diff
@@ -5,15 +5,15 @@
...
+	"description": "This function calls a given closure/function with every element in a given string list and returns true, if all of the closure calls returns true.",
 	"params": [
-		{"name": "list", "description": "", "required": true, "default": "", "type": "string", "values": []},
-		{"name": "closure", "description": "", "required": true, "default": "", "type": "function", "values": []},
-		{"name": "delimiter", "description": "", "required": false, "default": ",", "type": "string", "values": []},
-		{"name": "includeEmptyFields", "description": "", "required": false, "default": "false", "type": "boolean", "values": []},
-		{"name": "multiCharacterDelimiter", "description": "", "required": false, "default": "true", "type": "boolean", "values": []},
-		{"name": "parallel", "description": "", "required": false, "default": "false", "type": "boolean", "values": []},
-		{"name": "maxThreads", "description": "", "required": false, "default": "20", "type": "numeric", "values": []}
+		{"name": "list", "description": "String list to iterate.", "required": true, "default": "", "type": "string", "values": []},
```
shaedrich: I guess, you don't have to mention that lists are technically strings in Coldfusion.

data/en/listevery.json
```diff
@@ -5,15 +5,15 @@
...
+		{"name": "closure", "description": "Function/closure that implements the following constructor [function(any value, numeric index, string list, string delimiter):boolean].", "required": true, "default": "", "type": "function", "values": []},
```
shaedrich: We don't write closures in brackets in any other doc, I guess. I would call it "callback".

data/en/listevery.json
```diff
@@ -5,15 +5,15 @@
...
+		{"name": "delimiter", "description": "Delimiter used to separate the string list.", "required": false, "default": ",", "type": "string", "values": []},
+		{"name": "includeEmptyFields", "description": "Specifies whether to include empty fields or not.", "required": false, "default": "false", "type": "boolean", "values": []},
+		{"name": "multiCharacterDelimiter", "description": "Specifies whether the 'delimiter' parameter specifies a multi-character delimiter. If this parameter is true, the 'delimiter' parameter defines a single delimiter consisting of multiple characters.", "required": false, "default": "true", "type": "boolean", "values": []},
```
shaedrich: To be consistent I would reverse the statement: "wether the delimiter is a single-character delimiter. Than you don't have to mention that multi-character delimiter is the `false` value here.

data/en/listevery.json
SebastianZ: The docs are inconsistent regarding the wording. Sometimes they are called "closures", sometimes "callback", sometimes even different in the same description like in the [docs for `structEach()`](https://github.com/foundeo/cfdocs/blob/master/data/en/structeach.json). Even you [called them "closure" in a change for `structSort()` not long ago](https://github.com/foundeo/cfdocs/commit/f6c7eb40aa100c60ed4a78e194bde3dae86c5986#diff-f1fa85fa7507edae4fbf96256a34e7bbR14) and there is even a [guide calling callbacks "closures"](https://github.com/foundeo/cfdocs/blob/master/guides/en/closures.md). I'm happy to change the wording to "callback", though we should make sure we are consistent.

Also in regard of describing the signature of the callbacks the docs are inconsistent. Some of them define their parameters separately like [`arrayReduce()`](https://github.com/foundeo/cfdocs/blob/master/data/en/arrayreduce.json) and some display them in a syntactic way like [`replace()`](https://github.com/foundeo/cfdocs/blob/master/data/en/replace.json). So, what's the correct way to describe them?

Documentation rules like those two should be put into a separate document to avoid future inconsistencies.

data/en/listevery.json
SebastianZ: Wait, **`true`** is the value for the multi-character delimiter.

I'd even make it more precise, like "Specifies whether the 'delimiter' parameter is interpreted as a multi-character delimiter. If this parameter is true, the 'delimiter' parameter defines a single delimiter consisting of multiple characters, otherwise the characters are interpreted as multiple delimiters."

data/en/listevery.json
shaedrich: I know, these terms are mixed:
* function
* method (function in OOP)
* callback (function which is called after something is done)
* closure (named function)

> Documentation rules like those two should be put into a separate document to avoid future inconsistencies.

But that might be the question for a separate issue. I would really appreciate it if we could add the preferred nomenclature to the doc guide.

Using "callback" was just a suggestion, don't want to look like I want to enforce you to change it…

In my opinion the way in which `arrayReduce` does it, is more helpful.

data/en/listevery.json
shaedrich: That sounds confusing to me but this might happen because I'm not a native speaker. I would say something like this:

"If it is set to `true` all characters are considered as one delimiter; otherwise each character is a delimiter on its own."

SebastianZ:
I totally agree with you. The Lucee docs aren't in a perfect shape either. I just wanted to note from where I got the info about the function.

data/en/listevery.json
SebastianZ: > Using "callback" was just a suggestion, don't want to look like I want to enforce you to change it…

Callback is totally fine for me.

> In my opinion the way in which arrayReduce does it, is more helpful.

Sounds good. So I've changed that in https://github.com/foundeo/cfdocs/pull/871/commits/d6c0c6f588b8b965aa36e76e077772e7c8b2f8cb.

>> Documentation rules like those two should be put into a separate document to avoid future inconsistencies.

> But that might be the question for a separate issue. I would really appreciate it if we could add the preferred nomenclature to the doc guide.

Right. So I've created #876 to discuss this separately.

data/en/listevery.json
SebastianZ: Agree, that's easier to read. Changed it accordingly in https://github.com/foundeo/cfdocs/pull/871/commits/d6c0c6f588b8b965aa36e76e077772e7c8b2f8cb.

SebastianZ:
Looks like the build failed, though I'm not sure why. Travis says the tests ran into a timeout. Can the check be retriggered?

shaedrich:
Travis seems not to work that properly. I encountered the same problem yesterday. @pfreitag can retrigger it manually. To retrigger it by yourself you have to make another commit.

pfreitag:
Thanks @SebastianZ  and @shaedrich for reviewing

shaedrich:
My pleasure!

PR #874, opened 2018-06-27
Added examples to arrayEvery() function

SebastianZ:
Added one example that returns a positive result and one that returns a negative result.


pfreitag:
Thanks @SebastianZ

Issue #876, opened 2018-06-28
Create a style guide for the docs


SebastianZ:
Initially mentioned in https://github.com/foundeo/cfdocs/pull/871#discussion_r198614815, there should be a style guide defining some rules for how to write the docs.

This makes the different descriptions more consistent and complete and therefore easier to read.

For example, these are two things that this guide should clarify:

1. Which wording to use for callback functions like in [`structEach()`](https://github.com/foundeo/cfdocs/blob/master/data/en/structeach.json#L11). "callback", "closure", "closure / function reference", or something else?
2. How to describe the signature of callback functions. Inline like `function(param1, param2): boolean`, textual like in [`arrayReduce()`](https://github.com/foundeo/cfdocs/blob/master/data/en/arrayreduce.json#L25), a combination of both, or something different?

Sebastian

SebastianZ:
> closure (named function)

Note that "named function" is misleading, as they actually _not_ have a name. https://cfdocs.org/closures calls them "inner functions" and provides an example with an anonymous function being returned by a function.

shaedrich:
Turns out that it have to be both "named" and "inner". But it might be possible that these words were confused over the time or different languages implemented them differently.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures
https://stackoverflow.com/questions/4912116/closure-vs-anonymous-function-difference

Maybe there is another reliable source…

PR #877, opened 2018-06-28
Corrected and clarified description of getLocaleInfo() and added a third example

SebastianZ:
I've added the general description for the function, removed the incorrect `"discouraged"` item, fixed the name of the second parameter, explained the parameters in more detail, added a link to the Lucee documentation, added a third example showing how to change the output language, and improved the descriptions for the existing examples.

### DIFF
--- a/data/en/getlocaleinfo.json
+++ b/data/en/getlocaleinfo.json
@@ -4,29 +4,34 @@
 	"syntax":"getLocaleInfo()",
 	"returns":"struct",
 	"related":["getLocaleDisplayName","getLocale","getLocaleCountry","getLocaleLanguage"],
-	"description":"A short description that describes what the tag or function does.",
-	"discouraged":"Only add this key if this tag/function is discouraged by the community.",
+	"description":"Returns a structure containing info to a specific locale.\r\n\r\nThis function combines the locale functions getLocaleCountry, getLocaleDisplayName, and getLocaleLanguage to a single function.",
 	"params": [
-		{"name":"locale","description":"Geographic/language locale value","required":true,"default":"getLocale()","type":"string","values":["en","de_DE","de_ch","..."]},
-		{"name":"displayLocale","description":"Locale's display name/Output language","required":false,"default":"getLocaleDisplayName()","type":"string","values":["English","German","French","..."]}
+		{"name":"locale", "description": "Geographic/language locale value, where the format is a combination of an <a href=\"https://en.wikipedia.org/wiki/ISO_639-1\">ISO 639-1</a> code and an optional <a href=\"https://en.wikipedia.org/wiki/ISO_3166-1\">ISO 3166-1</a> code separated by a dash or an underscore.", "required": false, "default": "getLocale()", "type":"string", "values": ["en","de-DE","de-CH","..."]},
+		{"name":"dspLocale", "description": "Locale's display name/output language", "required": false, "default": "getLocaleDisplayName()", "type": "string", "values": ["English","German","French","..."]}
 	],
 	"engines": {
-		"lucee": {"minimum_version":"5", "notes":"Combines the locale functions <code>getLocaleCountry</code>, <code>getLocaleDisplayName</code>, <code>getLocaleInfo</code>, <code>getLocaleLanguage</code> to a single function", "docs":""}
+		"lucee": {"minimum_version": "5", "notes": "", "docs": "http://docs.lucee.org/reference/functions/getlocaleinfo.html"}
 	},
 	"links": [],
 	"examples": [
 		{
-				"title": "Get information about locale",
+				"title": "Get information about the page's locale",
 				"description": "",
 				"code": "getLocaleInfo()",
 				"result": "",
-				"runnable":false
-		},{
-				"title": "Output locale in a divergent language",
-				"description": "Same as above but different country display name output",
-				"code": "getLocaleInfo(getLocale(),'French').display.country",
+				"runnable": false
+		}, {
+				"title": "Output page's locale in a divergent language",
+				"description": "Outputs the locale of the page in French.",
+				"code": "getLocaleInfo(dspLocale='French').display.country",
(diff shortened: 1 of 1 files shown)


data/en/getlocaleinfo.json
```diff
@@ -4,29 +4,34 @@
...
-		{"name":"displayLocale","description":"Locale's display name/Output language","required":false,"default":"getLocaleDisplayName()","type":"string","values":["English","German","French","..."]}
+		{"name":"locale", "description": "Geographic/language locale value, where the format is a combination of an <a href=\"https://en.wikipedia.org/wiki/ISO_639-1\">ISO 639-1</a> code and an optional <a href=\"https://en.wikipedia.org/wiki/ISO_3166-1\">ISO 3166-1</a> code separated by a dash or an underscore.", "required": false, "default": "getLocale()", "type":"string", "values": ["en","de-DE","de-CH","..."]},
+		{"name":"dspLocale", "description": "Locale's display name/output language", "required": false, "default": "getLocaleDisplayName()", "type": "string", "values": ["English","German","French","..."]}
 	],
 	"engines": {
-		"lucee": {"minimum_version":"5", "notes":"Combines the locale functions <code>getLocaleCountry</code>, <code>getLocaleDisplayName</code>, <code>getLocaleInfo</code>, <code>getLocaleLanguage</code> to a single function", "docs":""}
```
shaedrich: Why is that wrong?

data/en/getlocaleinfo.json
```diff
@@ -4,29 +4,34 @@
...
 	"params": [
-		{"name":"locale","description":"Geographic/language locale value","required":true,"default":"getLocale()","type":"string","values":["en","de_DE","de_ch","..."]},
-		{"name":"displayLocale","description":"Locale's display name/Output language","required":false,"default":"getLocaleDisplayName()","type":"string","values":["English","German","French","..."]}
+		{"name":"locale", "description": "Geographic/language locale value, where the format is a combination of an <a href=\"https://en.wikipedia.org/wiki/ISO_639-1\">ISO 639-1</a> code and an optional <a href=\"https://en.wikipedia.org/wiki/ISO_3166-1\">ISO 3166-1</a> code separated by a dash or an underscore.", "required": false, "default": "getLocale()", "type":"string", "values": ["en","de-DE","de-CH","..."]},
+		{"name":"dspLocale", "description": "Locale's display name/output language", "required": false, "default": "getLocaleDisplayName()", "type": "string", "values": ["English","German","French","..."]}
```
shaedrich: Why is the "dsp-" prefix better?

SebastianZ: I used the "dsp" prefix because of http://docs.lucee.org/reference/functions/getlocaleinfo.html#argument-dspLocale (`displayLocale` is not recognized, I've checked).

And the lowercase "output" is because the sentence starts at "Locale's ...", not at "output", and because the previous description also has a lowercase "language" after the slash.
Actually, because the description is not a full sentence, it should be written all in lowercase. Though because the descriptions are often sentences, I still use a capital letter at the beginning.
Having said that, I'm not a native speaker nor grammar specialist. So I might be wrong about the official rules for that. Though this might be another case for #876, so I've added it there.

data/en/getlocaleinfo.json
SebastianZ: Three reasons:
1. I've moved that to the general description because I assume the functionality will be the same (combining the data of the other locale functions) when other engines implement it. This also has the nice side effect that the function names are linked to their descriptions automatically.
2. HTML isn't interpreted in engine notes, i.e. `<code>` is output as text. Might be worth another issue though.
3. The note contained `getLocaleInfo` itself, which doesn't make sense.

data/en/getlocaleinfo.json
shaedrich: The "output" is quite clear :+1:

Makes all sense. Good idea to add it there.

data/en/getlocaleinfo.json
shaedrich: It might be interpreted when it is markdown, but I'm not sure because not everything in the docs is parsed.

But you completely deleted the description oO

pfreitag: FYI You can use backticks instead of the `<code>` tag, but I don't think they are necessary for function names as it should link them automatically.

data/en/getlocaleinfo.json
```diff
@@ -4,29 +4,34 @@
...
-	"description":"A short description that describes what the tag or function does.",
-	"discouraged":"Only add this key if this tag/function is discouraged by the community.",
+	"description":"Returns a structure containing info to a specific locale.\r\n\r\nThis function combines the locale functions getLocaleCountry, getLocaleDisplayName, and getLocaleLanguage to a single function.",
 	"params": [
-		{"name":"locale","description":"Geographic/language locale value","required":true,"default":"getLocale()","type":"string","values":["en","de_DE","de_ch","..."]},
-		{"name":"displayLocale","description":"Locale's display name/Output language","required":false,"default":"getLocaleDisplayName()","type":"string","values":["English","German","French","..."]}
+		{"name":"locale", "description": "Geographic/language locale value, where the format is a combination of an <a href=\"https://en.wikipedia.org/wiki/ISO_639-1\">ISO 639-1</a> code and an optional <a href=\"https://en.wikipedia.org/wiki/ISO_3166-1\">ISO 3166-1</a> code separated by a dash or an underscore.", "required": false, "default": "getLocale()", "type":"string", "values": ["en","de-DE","de-CH","..."]},
```
pfreitag: Let's get rid of the HTML in here as well.

shaedrich: I'm not sure: Does it parse markdown in there? I guess, not.

So in which way can we link that or should we do an i18n guide?

pfreitag:
I fixed the html issue in commit 8514e043351d939f534c59b762ce8fdac5802a85

PR #879, opened 2018-06-28
Correct return type

### LINKED ISSUE #878: reFind Missing return values
This is missing the return values. I recommend adding that it returns the index of the match (1 or more) or if there is no match, returns 0.

shaedrich:
Resolve #878

### DIFF
--- a/data/en/refind.json
+++ b/data/en/refind.json
@@ -2,9 +2,9 @@
 	"name":"reFind",
 	"type":"function",
 	"syntax":"reFind(reg_expression, string [, start] [, returnsubexpressions])",
-	"returns":"number",
+	"returns":"any",
 	"related":["find","refindnocase","findnocase","rereplace"],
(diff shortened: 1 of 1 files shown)


KamasamaK:
The proper type name is `numeric`, not `number`.

The return type of `reFind` is only `numeric` when `returnsubexpressions` is false. When `returnsubexpressions` is true, it returns a `struct`. Therefore, `any` is the most apt to cover both of those cases.

shaedrich:
What about two signatures or a union type (`numeric | struct`)?

pfreitag:
I think `any` is most appropriate as the returntype - since it could return a struct in some cases as @KamasamaK pointed out. We don't have support for multiple return types right now so let's just make sure the description states that it will return `numeric` if `returnsubexpressions` is false and a `struct` of arrays named `len`, `match` and `pos` when `returnsubexpressions` is true.

Issue #774, opened 2018-01-16
Missing StructToSorted Function


pfreitag:
Adobe Docs: https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-s/structtosorted.html

Added to CF2016 update 3 https://helpx.adobe.com/coldfusion/kb/coldfusion-2016-update-3.html

KamasamaK:
@pfreitag This can be closed now

Issue #525, opened 2017-08-31
Syntax for functions and methods


shaedrich:
Now we have the keys `syntax` and `member` but `member` is definitely syntax as well. So that's a confusing naming, I guess. Would it be better to differentiate between `function_syntax` and `method_syntax`?

pfreitag:
I'm not sure if that would be more clear, maybe you can give me a more specific example of what it could look like.

shaedrich:
Something like that:
```json
{
    "name": "",
    "type": "",
    "syntax": [{
        "function": "ArrayReduce(arr, function() {})",
        "method": "arr.reduce(function() {})"
    }],
    ...
}
```

KamasamaK:
@shaedrich I think we should be very careful with changes like that which would break compatibility with the current format.

While we are discussing potential changes for representing member functions, I have a couple other ideas to throw out there.

1. Include member type. For the example of `arrayReduce` above, it should be `array`.
1. Include engine information with the same format as the entity itself and the same proposed for parameters.
1. Some distinct fields such as `member_name` would be nice instead of having to parse the `syntax` string.

shaedrich:
@KamasamaK You are right. But we could run a migration on the json files. Would be necessary when reworking the whole syntax anyway.

KamasamaK:
@shaedrich The migration isn't so much the issue as other tools using CFDocs. The Sublime Text plugin retrieves the JSON from the GitHub master repo, and parsing would break with a change like this. It would be good to get ahead of that and notify them of any breaking changes before they happen, but I wonder how many other projects or tools rely on such a process that we are unaware of.

shaedrich:
So we could change none of the recommended changes.

shaedrich:
Maybe we should launch another branch.

Are the data fetched by the api by most of the plugins unsing cfdocs?

This could be made better when provide an api versioned by [semver](http://semver.org).

KamasamaK:
The only plugin I am aware of that currently uses CFDocs is [sublimetext-cfml](https://github.com/jcberquist/sublimetext-cfml). A versioned API would absolutely solve this issue, but I was just bringing up the issue as to the current capabilities. If it's decided to go in the direction of making breaking changes, then a new branch would be appropriate and can be merged into `master` when it's deemed ready.

shaedrich:
[...]

PR #880, opened 2018-07-05
Add structToSorted function

aliaspooryorik:
Addresses: https://github.com/foundeo/cfdocs/issues/774

Think I've added it correctly to the listing .json files.


KamasamaK:
1. The names of the params should be consistent where possible. Right now, the first parameter is named `struct` and `anyStruct` in the `syntax` and `structure` in the `params` array.

1. `UDF` is not one of the [valid param types](https://github.com/foundeo/cfdocs/blob/master/data/cfdocs.schema.json#L99). Use `function`.

1. You can add a [`callback_params`](https://github.com/foundeo/cfdocs/blob/master/data/cfdocs.schema.json#L146) array within the callback's `params` entry to describe its params. Unfortunately, the cfdocs.org site does not yet read that, so it would be premature to remove it from `description`.

1. I see you added `structsetmetadata` to `data/en/struct-functions.json` but it appears to be missing from `data/en/all.json`. I'm not sure how that got through, but since you're modifying that file anyway, I figure you could add it.

aliaspooryorik:
@KamasamaK Thanks for feedback.

1. updated them all to `structure`
2. I copied the structeach.json file as my starting template and that used `UDF` so I've updated that config to `function` as well as the structtosorted.json.
3. Not quite sure what you mean here - do you have an example?
4. `structsetmetadata` added to all.json as requested

KamasamaK:
Regarding 3, I don't think there are any examples yet in the JSON data, but [here is the closest I have to an example](https://github.com/foundeo/cfdocs/issues/525#issuecomment-327285363). Just know that schematically, `callback_params` and `params` are the same. This would basically be what you put into the param `description` transposed to `callback_params`:

```json
"callback_params": [
	{
			"name": "value1",
			"description": "Value of the key/value pair in the struct",
			"required": true,
			"default": "",
			"type": "any",
			"values": []
	},
	{
			"name": "value2",
			"description": "Value of the key/value pair to compare against",
			"required": true,
			"default": "",
			"type": "any",
			"values": []
	},
	{
			"name": "key1",
			"description": "Key name of the key/value pair in the struct",
			"required": true,
			"default": "",
			"type": "string",
			"values": []
	},
	{
			"name": "key2",
			"description": "Key name of the key/value pair to compare against",
			"required": true,
			"default": "",
			"type": "string",
			"values": []
	}
]
```

KamasamaK:
Since JSON schema validation is case-sensitive, make sure the [data types match exactly](https://github.com/foundeo/cfdocs/blob/35c75c11113d5eb25494ce8542c1c4649544db38/data/cfdocs.schema.json#L99). e.g. `Any` vs. `any`. Looks good otherwise.

@pfreitag What are your thoughts on including HTML in the data? e.g. `<strong>`

aliaspooryorik:
Have updated to `any`. I got the `Any` case from the linked example: https://github.com/foundeo/cfdocs/issues/525#issuecomment-327285363

As for HTML in the data - I got that from StructEach doc which I used as a template: 
https://github.com/foundeo/cfdocs/blob/master/data/en/structeach.json#L11

pfreitag:
@KamasamaK - I don't want HTML in the JSON.

The site currently supports back ticks in most places which causes the text to be wrapped in a `<code>` tag. 

If we really need strong then I'd favor using `*something bold*` as the syntax instead of HTML. I'd welcome your opinion as you are a plugin developer using this repository - thoughts?

pfreitag:
@aliaspooryorik thanks for your contributions here, I'm going to merge this PR - we can remove the HTML in another PR (I know you didn't add it in this one you were just updating what was there).

KamasamaK:
For the usage here, I agree that backticks would be a better alternative. Incidentally, the platform I work on uses a subset of markdown, so the backticks work well there. I've created #883 for further discussion.

PR #890, opened 2018-07-20
New ColdFusion 2018 functions

### LINKED ISSUE #887: New ColdFusion 2018 functions
The following functions are new to ColdFusion 2018. Several were already supported by Lucee, so the engine information just needs to be updated.

- [arrayFirst](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-a-b/arrayfirst.html) (this already exists for Lucee)
- [arrayLast](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-a-b/arraylast.html) (this already exists for Lucee)
- cacheGetEngineProperties (no doc found): Returns information about the [cache engine](https://helpx.adobe.com/coldfusion/developing-applications/changes-in-coldfusion/caching-enhancements.html)
- [queryDeleteColumn](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-m-r/querydeletecolumn.html) (this already exists for Lucee)
- [queryDeleteRow](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-m-r/querydeleterow.html) (this already exists for Lucee)
- [runAsync](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-m-r/runasync.html)
- [structIsOrdered](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-s/structisordered.html)

shaedrich:
Resolve #887


data/en/runasync.json
```diff
@@ -0,0 +1,46 @@
+{
+    "name":"runAsync",
+    "type":"function",
+    "syntax":"runAsync(function() {})",
+    "script":"runAsync(function() {});",
+    "returns":"Future",
```
KamasamaK: I do not see that they added this as a new return type. This should probably be `any`.

shaedrich: They say, it is called `Future` which is something like a `Promise` in JavaScript. Should we really use something unspecific like `any`?

ddspringle: @shaedrich To be clear, `Future` in CFML is nothing like a `Promise` in JavaScript (its a subtle difference, but is different enough not to confuse the two). It is closer to a `CompletableFuture` in Java. 

That said, and in this case, I think `any` is probably appropriate here given it returns any value... it just happens to return it at a future point in the code (which could perhaps be spelled out in the description, if it isn't already)

pfreitag:
Nice work - thanks!

shaedrich:
My pleasure!

KamasamaK:
I still don't think it makes sense to use `script` for anything but tags.

pfreitag:
@KamasamaK - not sure what you were referring to here?

pfreitag:
Ah yes - I agree - that is only necessary for tags to show the cfscript equivalent.

Issue #812, opened 2018-03-06
Polyfills


shaedrich:
Would it be a good idea to provide polyfills for engine versions without support for a specific function or tag?

To migrate this, we could use [these overrides](https://gist.github.com/anonymous/490abe8522804f754dbf3456efdc7d54) from [watney](/ryanguill/watney/) by [@ryanguill](/ryanguill).

pfreitag:
I think these could be added as examples

shaedrich:
To easier distinguish polyfills from examples, "code sections" could be tagged. There are probably more types than just "example" or "polyfill":
* sample usage
* polyfill
* exceptions
* configuration
* etc.

That could also help with grouping, sorting and perhaps hiding "code sections"

PR #904, opened 2018-08-16
Add member function to structAppend()

shaedrich:


data/en/structappend.json
```diff
@@ -2,13 +2,14 @@
 	"name":"structAppend",
 	"type":"function",
 	"syntax":"structAppend(destStruct, sourceStruct [, overwriteFlag])",
+	"member": "destStruct.append(sourceStruct [, overwriteFlag])\ndestStruct.putAll(sourceStruct [, overwriteFlag])",
```
KamasamaK: `putAll` is not a member function. It works because it is falling back on Java's `putAll` method for `Map`, which does not even support the optional second parameter and fails if provided.

shaedrich: Oh, okay. Thanks for the hint. I guess we shouldn't refer to java functions doing mostly the same?

KamasamaK: Since it's not part of CFML and a result of implementation, I would say that it should not be included. It seems that @pfreitag has already merged this PR, but I would be interested in his take on this.

pfreitag: Yeah - I'd tend to agree that the java method is technically _undocumented_ and just a side effect of the implementation. I think it would be more suitable as an example to show how you could do it in earlier versions of CF.

KamasamaK: So kind of related to #812.

shaedrich: Sounds good :+1:

So, I make a new pr related to this one with the requested example, right?

PR #915, opened 2018-08-29
Update cfqueryparam.json

rnallapati:
"null" and "list" attributes have a possible values of "yes" or "no" (and not "true" or "false")


shaedrich:
@pfreitag Do you think, it's good to use `true` and `false` for truthy and falsy values in `values` param but `YES` and `NO` in description so that we're inconsistent?

PR #917, opened 2018-09-04
Remove duplicated params in param description and improved function d…

shaedrich:
…escription


shaedrich:
@pfreitag Bug which caused checks to fail was fixed in #919.

PR #918, opened 2018-09-04
Reorder param struct content & add callback_params

shaedrich:


shaedrich:
@pfreitag Bug which caused checks to fail was fixed in #919.

PR #919, opened 2018-09-04
Fix wrong relation

shaedrich:
> * (!)JSON Related Tests 
>    * (!)should relate to an existing function or tag (157 ms) 
>	    * -> Failure: encodeForCN is/are non-existing function(s)/tag(s) in cfldap


pfreitag:
Thanks -- I goofed that up.

shaedrich:
You're welcome. It's good that we notice that at all.

PR #920, opened 2018-09-04
expand explanation for `to` attribute

bardware:
Add most sought for information; taken from [Adobe](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-m-o/cfmail.html)

### DIFF
--- a/data/en/cfmail.json
+++ b/data/en/cfmail.json
@@ -246,7 +246,7 @@
 			"name": "to",
 			"required": true,
 			"type": "string",
-			"description": "Message recipient email addresses.",
+			"description": "Message recipient email addresses. To specify multiple addresses, separate the addresses with commas.",
 			"default": "",
 			"values": []
 		},


data/en/cfmail.json
```diff
@@ -246,7 +246,7 @@
 			"name": "to",
 			"required": true,
 			"type": "string",
-			"description": "Message recipient email addresses.",
+			"description": "Message recipient email addresses. To specify multiple addresses, separate the addresses with commas.",
```
shaedrich: I'm not sure if we should just say, a list should be used in that case.

pfreitag: I think comma is good, because a _list_ does not necessarily have a comma delimiter. There is no way to specify a different delimiter right?

shaedrich: So I would suggest saying "comma separated list".

bardware: Thanks for accepting the merge request.
I saw no documentation that the list of recipients is passed to the mailserver. Only read rumors. I'd like to believe Adobe is right. I remember looking up if I should use comma or semicolon repeatedly in the past.

PR #925, opened 2018-09-20
Bug notice better in engine notes

shaedrich:


KamasamaK:
The Lucee bug is https://luceeserver.atlassian.net/browse/LDEV-574 and shows being fixed in 5.2.7

PR #931, opened 2018-10-09
Added note in CF engine that QueryFilter mutates the original query

ckarras11:
I added the bug report and description.  VS Code formatted the file on save and spaced some of the JSON out.  If that is an issue I can revert the changes, turn off autoformatter and commit again.  Thanks!


pfreitag:
Thanks @ckarras11

PR #932, opened 2018-10-10
Update index files

cfvonner:
Running the /utilities/indexer.cfm added some new function entries for various versions.


KamasamaK:
I see that `arrayGetMetadata`, `arraySetMetadata`, `structGetMetadata` and `structSetMetadata` were removed when this was run. Should functions added in minor updates not be included or is this a bug?

cfvonner:
I didn't catch that. We may need standalone documents for the update
versions to match the version tags.

------
Sent from my mobile device.

Carl Von Stetten
carl.vonstetten1@gmail.com
http://about.me/cfvonner

cfvonner:
Incidentally, I didn't mean for that second commit to get added to the same
pull request. Not sure how to split it off now.

------
Sent from my mobile device.

Carl Von Stetten
carl.vonstetten1@gmail.com
http://about.me/cfvonner

cfvonner:
Why don't you hold back on this PR for a day or two. Ill see if I can add
those updates to the indexer.cfm logic.

------
Sent from my mobile device.

Carl Von Stetten
carl.vonstetten1@gmail.com
http://about.me/cfvonner

cfvonner:
I have a fix in the works, but there's one piece I need to get insight on.  At the end of line 26 of indexer.cfm ( [https://github.com/foundeo/cfdocs/blob/cd1b3b1e3da7f4dc18858cbbe67f78e0dfd6cefa/utilities/indexer.cfm#L26](url) ), there is this test: 
`isNumeric(data.engines.coldfusion.minimum_version)`.

This test will allow complete indexes to be built for those update releases if we refer to the ColdFusion updater versions as `2016.2` instead of the more semantically correct `2016.0.2`.  Is this tradeoff worthwhile or is it perhaps misleading?  Otherwise we'd need to remove the `isNumeric` test.

shaedrich:
Well, most of the time we use just the major version. If we would want to use a semantic version (If I remember right, we don't want to) we should have a small functions that checks the version according to semver syntax…

Maybe a question for @pfreitag and/or @KamasamaK

cfvonner:
@shaedrich The issue was that functions introduced in updates (not in the initial release) had the `engines.coldfusion.minimum_version` in their respective json file set to `2016.0.2` or similar.  When the indexer.cfm was run, these functions were removed from the `cf2016.json` index file and didn't land in the newly created `cf201602.json` index file.  By changing the `engines.coldfusion.minimum_version` to `2016.2` in those function json files, the indexer created a `cf20162.json` file and had those functions properly included.

pfreitag:
Yeah I see your point, with respect to the indexer it creates a file like cf2016.json which should show all new functions added in CF2016. I think that file should also include any functions added in any release of CF2016. So there is probably a bug in the indexer that needs to be fixed. It should use the versions array that is defined on line 8 (or line 9 for lucee) to create the version index files. 

As for semver, I think we had a discussion about this at one point, I don't recall if there was a reason not to use semver - I can't think of any reason to use `2016.2` instead of `2016.0.2` maybe there is another place it is treating the version number as numeric (maybe the json schema file?)

cfvonner:
I can modify the logic between line 26 and 31 to simplify the `minimum_version` down to just the major release for the purposes of building up the version-specific function/tag array.  Basically just strip off the trailing `.0.2` from the version.

shaedrich:
I would suggest using a separate pr for this.

cfvonner:
I will do a separate pull request and close this one.  I'll also get the closures.md updates in as a separate pull request.

cfvonner:
I've got the logic worked out and am getting ready to submit a new PR.  Should the change to `indexer.cfm` be in a separate PR from the resulting updated index json files?  Also, I had to make one minor correction to the semantic version number in `getCononicalPath.json` to get it to index properly - should that be a separate PR as well?  Or is this all collectively worthy of a single PR?

pfreitag:
Go for separate PRs, incase there is an issue with one of the things you can iterate on that one thing, instead of holding up the rest of the changes. Also hacktoberfest 👍

Issue #883, opened 2018-07-06
Formatting the data


KamasamaK:
See [PR 880 comment](https://github.com/foundeo/cfdocs/pull/880#issuecomment-403067817) and [PR 877 comment thread](https://github.com/foundeo/cfdocs/pull/877#pullrequestreview-133343582) for background.

I agree that HTML should be avoided. I am curious if anyone thinks more markdown should be supported or if we should we avoid more formatting. Regardless, it should be documented what's supported.

I've also found that there are too many new lines in places they are not needed, such as in the middle of a sentence. I can understand in some cases wanting to start a new line, but I do not think, for instance, that we should be enforcing some kind of line length in the data.

Please provide your opinions on this.

pfreitag:
My thoughts on this are:

* I don't want HTML in the json
* I think a subset of markdown might make sense (eg backticks are already supported), but I don't think full blown markdown is necessary. My question would be what formatting options of markdown do we need and why? My concern with a full markdown implementation is that the docs might be harder for third parties to use (eg IDE's), it's not really a technical one since we already support markdown for the guides.
* I agree that the newlines should be removed, I have been removing them when I edit a doc that has them.

shaedrich:
We should have an overview of which formatting is supported for which values.

shawnoden:
Did we ever build out this formatting overview? A lot of what I've been doing is trying to correct formatting on the tags when I come across them, but I didn't really see any consistent documentation of preferred formatting.

shaedrich:
I don't think so. But feel free to start that.

pfreitag:
We'll let's start with my question from a few comments ago: _My question would be what formatting options of markdown do we need and why?_  

As noted above backticks are currently supported, but if you get too complicated with markdown support then different markdown engines have slightly different results (eg: some support tables some do not).  

If we can get away with a subset of full blown markdown support I think that would be best. Especially for those supporting the descriptions in IDE's or other projects it complicates their requirements (you don't need a markdown library to replace back ticks with `<code>` tags but you may need a library for more complicated markdown features).

shawnoden:
[...]

PR #935, opened 2018-10-10
Added descriptions to arguments lsdayofweek

Miguel-F:
Added descriptions to arguments from the Lucee documentation located here - https://docs.lucee.org/reference/functions/lsdayofweek.html

NOTE: I added some HTML tabs within the description. I am assuming that will work.  If not then just reject this change and I will modify.


shaedrich:
Do you want to achieve a line break by using `<p />`? It's easier to use `\n` than wrapping it in a tab.

I'm not sure, but I guess, param descriptions are not parsed so this might not work.

Miguel-F:
Thanks for the feedback.  I wasn't really sure what was allowed there.  I was also trying to create a bulleted list.  Is there any option for that?

shaedrich:
People use either "*" or "-" and delimit them using `\n`

In general this could be interesting to read:
* https://github.com/foundeo/cfdocs/issues/883

Somewhere we had a discussion about either to use "*" or "-" but I can't find it right now.

Miguel-F:
I removed the HTML tags and replaced them with `\n` to avoid any potential issues.

pfreitag:
Thanks for contributing and fixing @Miguel-F and thanks for reviewing @shaedrich

shaedrich:
My pleasure, @pfreitag!

PR #936, opened 2018-10-10
Added descriptions to arguments for millisecond

Miguel-F:
Added descriptions to arguments from the Lucee documentation located here - https://docs.lucee.org/reference/functions/millisecond.html

### DIFF
--- a/data/en/millisecond.json
+++ b/data/en/millisecond.json
@@ -5,10 +5,10 @@
 	"member": "date.millisecond([timezone])",
 	"returns": "numeric",
 	"related": [],
-	"description": "",
+	"description": "Returns the milliseconds of the referenced date.",
 	"params": [
-		{"name": "date", "description": "", "required": true, "default": "", "type": "date", "values": []},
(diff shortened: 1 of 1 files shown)


data/en/millisecond.json
```diff
@@ -5,14 +5,14 @@
 	"member": "date.millisecond([timezone])",
 	"returns": "numeric",
 	"related": [],
-	"description": "",
+	"description": "Returns the milliseconds of the referenced date.",
 	"params": [
-		{"name": "date", "description": "", "required": true, "default": "", "type": "date", "values": []},
-		{"name": "timezone", "description": "", "required": false, "default": "", "type": "string", "values": []}
+		{"name": "date", "description": "date object; for example, now()", "required": true, "default": "", "type": "date", "values": []},
```
shaedrich: I'm not sure if providing an example is really needed in this place…

pfreitag: Probably not necessary, but it's not too wordy so I'm fine with it.

PR #937, opened 2018-10-10
Added descriptions to arguments for lsweek

Miguel-F:
Added descriptions to arguments from the Lucee documentation located here - https://docs.lucee.org/reference/functions/lsweek.html

NOTE: I added some HTML tabs within the description. I am assuming that will work. If not then just reject this change and I will modify.


data/en/lsweek.json
```diff
@@ -4,15 +4,15 @@
...
 	"returns": "numeric",
 	"related": [],
-	"description": "",
+	"description": "Returns the week number in the year of the date referenced, based on the local influence.\n\nNot everywhere does the week begin on Monday by definition.",
 	"params": [
-		{"name": "date", "description": "", "required": true, "default": "", "type": "date", "values": []},
-		{"name": "locale", "description": "", "required": false, "default": "", "type": "string", "values": []},
-		{"name": "timezone", "description": "", "required": false, "default": "", "type": "string", "values": []}
+		{"name": "date", "description": "date object; for example, now()", "required": true, "default": "", "type": "date", "values": []},
+		{"name": "locale", "description": "Locale to use instead of the locale of the page when processing the function", "required": false, "default": "", "type": "string", "values": []},
```
shaedrich: I don't know if we have to note that the page's locale is the default here if `locale` is not set…

Miguel-F:
I removed the HTML tags and replaced them with `\n` to avoid any potential issues.

PR #939, opened 2018-10-10
Added description and parameters

alorne:


data/en/gettimezone.json
```diff
@@ -4,8 +4,8 @@
 	"syntax": "getTimezone()",
 	"returns": "string",
 	"related": ["getTimezoneInfo","setTimezone","getLocale"],
-	"description": "",
-	"params": [],
+	"description": "Returns the timezone for the current request.",
+	"params": ["This function does not take any arguments"],
```
shaedrich: Please don't do that. This will break the doc.

alorne: I'm sorry, I'm new at this. What exactly is wrong?

shaedrich: No worries! `params` is either an empty array or a array of structs. So if you'd pass a string it considers it an non-empty array, tries to loop over the structs and wonders why it gets a simple value.

alorne: Ok, I'll fix it. Thanks!

shaedrich: Thanks! You're welcome!

pfreitag:
Thanks!

PR #940, opened 2018-10-10
Added description to function for imagefonts

Miguel-F:
Added description to function from the Lucee documentation located here - https://docs.lucee.org/reference/functions/imagefonts.html

### DIFF
--- a/data/en/imagefonts.json
+++ b/data/en/imagefonts.json
@@ -4,10 +4,10 @@
 	"syntax": "imageFonts()",
 	"returns": "array",
 	"related": [],
-	"description": "",
+	"description": "return all available",
 	"params": [],
 	"engines": {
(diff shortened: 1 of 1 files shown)


data/en/imagefonts.json
```diff
@@ -4,11 +4,11 @@
 	"syntax": "imageFonts()",
 	"returns": "array",
 	"related": [],
-	"description": "",
+	"description": "return all available",
```
shaedrich: All available what?

Miguel-F: All available image fonts.  Right?  I pulled that from the Lucee documentation.

shaedrich: yup. If you change it to "Returns all available image fonts", it's fine ;-)

PR #941, opened 2018-10-10
Added description to function for gettimezone

Miguel-F:
Added descrition for the function gettimezone from the Lucee documentation located here - https://docs.lucee.org/reference/functions/gettimezone.html


Miguel-F:
Okay, my mistake.  I will remove this pull request.

PR #942, opened 2018-10-10
Added description to function for ESAPIDecode

Miguel-F:
Added description to function from the Lucee documentation located here - https://docs.lucee.org/reference/functions/esapidecode.html

### DIFF
--- a/data/en/esapidecode.json
+++ b/data/en/esapidecode.json
@@ -4,10 +4,10 @@
 	"syntax": "esapiDecode(decodeFrom, string)",
 	"returns": "string",
 	"related": ["esapiEncode"],
-	"description": "",
+	"description": "Decodes a string that has been encoded with ESAPIEncode.",
 	"params": [
-		{"name": "decodeFrom", "description": "", "required": true, "default": "", "type": "string", "values": []},
-		{"name": "string", "description": "", "required": true, "default": "", "type": "string", "values": []}
+		{"name": "decodeFrom", "description": "given value is encoded as:\n\n* url: also decodes html-encoded content in lucee 4\n* html (not in lucee 4)", "required": true, "default": "", "type": "string", "values": []},
+		{"name": "string", "description": "string to encode", "required": true, "default": "", "type": "string", "values": []}
 	],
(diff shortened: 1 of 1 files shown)


data/en/esapidecode.json
```diff
@@ -4,10 +4,10 @@
 	"syntax": "esapiDecode(decodeFrom, string)",
 	"returns": "string",
 	"related": ["esapiEncode"],
-	"description": "",
+	"description": "Decodes a string that has been encoded with ESAPIEncode.",
 	"params": [
-		{"name": "decodeFrom", "description": "", "required": true, "default": "", "type": "string", "values": []},
-		{"name": "string", "description": "", "required": true, "default": "", "type": "string", "values": []}
+		{"name": "decodeFrom", "description": "given value is encoded as:\n\n* url: also decodes html-encoded content in lucee 4\n* html (not in lucee 4)", "required": true, "default": "", "type": "string", "values": []},
```
shaedrich: Instead of trailing "(not in Lucee 4)" you should write leading Lucee5+ and corresponding CFx+

Miguel-F: As I understand it, the version information is specific to each value of the `decodeFrom` parameter.  The `url` value can also decode html-encoded content from lucee 4 but the `html` value cannot. The version information does not pertain to the function itself, just these values.

shaedrich: That's correct. You can specify the engine version by param. There is no `mininum_version` like in `engines` but you can say `"description": "CF2016+ Lucee5+ some text"` for example which will be parsed to a label.

Miguel-F: Ah, okay.  Thanks for the info.

shaedrich: You're welcome!

Miguel-F: I have updated this pull request with your suggestion. Notice that I added the labels in front of each value (in the middle of the description). Will it still parse those correctly?

shaedrich: Looks good. No worries, the will :wink:

PR #946, opened 2018-10-10
Added descriptions to arguments for listTrim and deprecated

Miguel-F:
Added descriptions to arguments from the Lucee documentation located here - https://docs.lucee.org/reference/functions/listtrim.html
And added deprecation note according to those docs and here https://github.com/lucee/Lucee/blob/5.2.9.31/core/src/main/java/lucee/runtime/functions/list/ListTrim.java#L36

### DIFF
--- a/data/en/listtrim.json
+++ b/data/en/listtrim.json
@@ -10,7 +10,7 @@
 		{"name": "delimiters", "description": "characters that separate list elements", "required": false, "default": ",", "type": "string", "values": []}
 	],
 	"engines": {
-		"lucee": {"minimum_version": "", "deprecated": 4, "notes": "use ListCompact instead", "docs": "http://docs.lucee.org/reference/functions/listtrim.html"}
+		"lucee": {"minimum_version": "", "deprecated": "4", "notes": "use ListCompact instead", "docs": "http://docs.lucee.org/reference/functions/listtrim.html"}
 	},
 	"examples": [],
 	"links": []
--- a/data/en/listtrim.json
+++ b/data/en/listtrim.json
@@ -7,7 +7,7 @@
 	"description": "trims leading and trailing spaces from each list element",
 	"params": [
 		{"name": "list", "description": "a string list", "required": true, "default": "", "type": "string", "values": []},
-		{"name": "delimiters", "description": "characters that separate list elements. The default value is comma.", "required": false, "default": ",", "type": "string", "values": []}
+		{"name": "delimiters", "description": "characters that separate list elements", "required": false, "default": ",", "type": "string", "values": []}
 	],
 	"engines": {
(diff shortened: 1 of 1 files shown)


data/en/listtrim.json
```diff
@@ -4,14 +4,14 @@
 	"syntax": "listTrim(list [, delimiters])",
 	"returns": "string",
 	"related": [],
-	"description": "",
+	"description": "trims leading and trailing spaces from each list element",
 	"params": [
-		{"name": "list", "description": "", "required": true, "default": "", "type": "string", "values": []},
-		{"name": "delimiters", "description": "", "required": false, "default": ",", "type": "string", "values": []}
+		{"name": "list", "description": "a string list", "required": true, "default": "", "type": "string", "values": []},
+		{"name": "delimiters", "description": "characters that separate list elements. The default value is comma.", "required": false, "default": ",", "type": "string", "values": []}
```
shaedrich: You don't need the second sentence ("The default value is comma") if you specified the default value in the structure.

Miguel-F:
okay I removed that sentence and updated the pull request

data/en/listtrim.json
```diff
@@ -4,14 +4,14 @@
...
+	"description": "trims leading and trailing spaces from each list element",
 	"params": [
-		{"name": "list", "description": "", "required": true, "default": "", "type": "string", "values": []},
-		{"name": "delimiters", "description": "", "required": false, "default": ",", "type": "string", "values": []}
+		{"name": "list", "description": "a string list", "required": true, "default": "", "type": "string", "values": []},
+		{"name": "delimiters", "description": "characters that separate list elements", "required": false, "default": ",", "type": "string", "values": []}
 	],
 	"engines": {
-		"lucee": {"minimum_version": "", "notes": "", "docs": "http://docs.lucee.org/reference/functions/listtrim.html"}
+		"lucee": {"minimum_version": "", "deprecated": 4, "notes": "use ListCompact instead", "docs": "http://docs.lucee.org/reference/functions/listtrim.html"}
```
KamasamaK: Version properties such as `deprecated` should have string values.

Miguel-F: I copied that from pull request #945  The value is not a string in that change?

shaedrich: Mea culpa!

Miguel-F: Ha Ha.  No problem. I updated the pull request.

KamasamaK: You should not consider unmerged PRs to be indicative of what is correct. There are many existing data files with this property that are all strings. You could also consider that it would just make sense for it to be the same as for any other version property (see `minimum_version`). I would ultimately advise you just look at the [schema](https://github.com/foundeo/cfdocs/blob/0e28a0793e734279cba49af8452b544d4b8beb0f/data/cfdocs.schema.json#L14) if you have any questions or doubts.

Miguel-F: Understood.

shaedrich: Okay, you're right.

PR #951, opened 2018-10-10
Add support for semantic version numers to listing indexes

cfvonner:
Puts tags/functions added in an update version into the list for the major release.

Will handle current versioning (2016.0.#) and past versioning (11.0.#).

### DIFF
--- a/utilities/indexer.cfm
+++ b/utilities/indexer.cfm
@@ -23,10 +23,16 @@
 			<cfelseif data.type IS "listing">
 				<cfset arrayAppend(categories, nameKey)>
 			</cfif>
-			<cfif structKeyExists(data, "engines") AND structKeyExists(data.engines, "coldfusion") AND structKeyExists(data.engines.coldfusion, "minimum_version") AND isNumeric(data.engines.coldfusion.minimum_version)>
-
-				<cfif structKeyExists(versions, val(data.engines.coldfusion.minimum_version))>
-					<cfset arrayAppend(versions[val(data.engines.coldfusion.minimum_version)], data.name)>
+			<cfif structKeyExists(data, "engines") AND structKeyExists(data.engines, "coldfusion") AND structKeyExists(data.engines.coldfusion, "minimum_version")>
+				<cfif isNumeric(data.engines.coldfusion.minimum_version)>
+					<cfset minimum_version = data.engines.coldfusion.minimum_version>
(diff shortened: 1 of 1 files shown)


utilities/indexer.cfm
```diff
@@ -23,10 +23,16 @@
...
 				<cfset arrayAppend(categories, nameKey)>
 			</cfif>
-			<cfif structKeyExists(data, "engines") AND structKeyExists(data.engines, "coldfusion") AND structKeyExists(data.engines.coldfusion, "minimum_version") AND isNumeric(data.engines.coldfusion.minimum_version)>
-
-				<cfif structKeyExists(versions, val(data.engines.coldfusion.minimum_version))>
-					<cfset arrayAppend(versions[val(data.engines.coldfusion.minimum_version)], data.name)>
+			<cfif structKeyExists(data, "engines") AND structKeyExists(data.engines, "coldfusion") AND structKeyExists(data.engines.coldfusion, "minimum_version")>
+				<cfif isNumeric(data.engines.coldfusion.minimum_version)>
+					<cfset minimum_version = data.engines.coldfusion.minimum_version>
+				<cfelseif reFind('^[0-9]{2,4}\.[0-9]\.[0-9]+$', data.engines.coldfusion.minimum_version)>
```
shaedrich: :+1: :heart:

utilities/indexer.cfm
```diff
@@ -23,10 +23,16 @@
...
+					<cfset minimum_version = reFind( '^([0-9]{2,4})', data.engines.coldfusion.minimum_version, 1, true ).match[1]>
+				<cfelse>
+					<cfset minimum_version = ''>
+				</cfif>
+				<cfif structKeyExists(versions, val(minimum_version))>
```
shaedrich: Not your fault but a note to old code: Railo has version 0. If you set `minimum_version` to `''` and run it through `val()`, you'll get 0 which might exist in versions struct but currently does not so there was not unexpected behavoir yet.

cfvonner: I didn't know that! 👍  

What do you think would be a better placeholder for non-compliant version numbers?  `-1` maybe?

shaedrich: Or just don't set it and then check for existence.

cfvonner: I'll work on an update to this tomorrow.  I like the idea of not setting it and checking for existence.

PR #955, opened 2018-10-11
Update createguid.json

CraicOverflow89:
Added simple example.

### DIFF
--- a/data/en/createguid.json
+++ b/data/en/createguid.json
@@ -10,5 +10,13 @@
 		"lucee": {"minimum_version": "", "notes": "", "docs": "http://docs.lucee.org/reference/functions/createguid.html"}
 	},
 	"links": [],
-	"examples": []
+	"examples": [
+		{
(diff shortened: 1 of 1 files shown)


data/en/createguid.json
```diff
@@ -10,5 +10,13 @@
 		"lucee": {"minimum_version": "", "notes": "", "docs": "http://docs.lucee.org/reference/functions/createguid.html"}
 	},
 	"links": [],
-	"examples": []
+	"examples": [
+		{
+			"title":"Creates a Globally Unique Identifier (GUID). Each GUID will be unique",
+			"description":"",
+			"code":"createGUID()",
```
shaedrich: You could prove that it is always unique by something like `createGUID() neq createGUID()`.

pfreitag: Lets do that as a second example (feel free to do that as another PR @CraicOverflow89 )... I think people may want to run that one just to see what the guid looks like.

PR #956, opened 2018-10-11
Update fileopen.json

CraicOverflow89:
Added example using fileOpen, fileReadLine and fileClose for standard usage.

### DIFF
--- a/data/en/fileopen.json
+++ b/data/en/fileopen.json
@@ -21,7 +21,7 @@
 		{
 			"title":"Opens a file, reads a line then closes it.",
 			"description":"",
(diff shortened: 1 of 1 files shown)


data/en/fileopen.json
```diff
@@ -17,5 +17,13 @@
...
 	},
 	"links": [],
-	"examples": []
+	"examples": [
+		{
+			"title":"Opens a file, reads a line then closes it.",
+			"description":"",
+			"code":"// Open File\nvar fileObject = fileOpen(\"/path/to/file.txt\");\n\n// Perform Actions\ntry\n{\n\t// Read Line\n\twriteOutput(fileReadLine(fileObject));\n}\n\n// Error Handling\ncatch(any ex)\n{\n\t// Report Exception\n\twriteDump(ex);\n}\n\n// Always Close\nfinally\n{\n\t// Close File\n\tfileClose(fileObject);\n}",
+			"result":"",
+			"runnable":true
```
shaedrich: file methods are disabled on trycf so running that example will cause an exception.

pfreitag: Yup - can you update the runnable flag to be false, that will remove the run button.

PR #960, opened 2018-10-11
Fixed listFilter example

ckarras11:
listFilter example was copied from arrayFilter.  Changed to listFilter with a simple numbers list.  Also removed non-existent member function.

### DIFF
--- a/data/en/listfilter.json
+++ b/data/en/listfilter.json
@@ -1,42 +1,59 @@
 {
-	"name":"listFilter",
-	"type":"function",
-	"syntax":"listFilter(list, function(listElement, [list]) )",
-	"member":"someList.filter(function(listElement, [list]) )",
-	"returns":"string",
-	"related":["listGetAt"],
-	"description":"Used to filter an list to items for which the closure function returns true.",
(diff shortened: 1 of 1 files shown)


data/en/listfilter.json
```diff
@@ -1,42 +1,59 @@
...
-	"type":"function",
-	"syntax":"listFilter(list, function(listElement, [list]) )",
-	"member":"someList.filter(function(listElement, [list]) )",
-	"returns":"string",
-	"related":["listGetAt"],
-	"description":"Used to filter an list to items for which the closure function returns true.",
+	"name": "listFilter",
+	"type": "function",
+	"syntax": "listFilter(list, function(listElement, [list]) )",
+	"member": "someList.filter(function(listElement, [list]) )",
```
shaedrich: For some reason the member function is `string.listFilter()`

ckarras11: I will change that member function as well as make an example for it

pfreitag: Ok - you can create another PR to make those changes. I'll merge this one.

PR #963, opened 2018-10-12
Expand the first five examples by default

### LINKED ISSUE #864: Examples area
I just wanted to voice my opinion about the examples area. 

Having all the syntax closed by default is inconvenient especially when all the headings are the same with no clue as to what it is going to uncover. 
Also only allowed to have one open at a time is even more so. Now you can't compare one over another over another.

image

shaedrich:
and let them collapse/expand independently
Fix #864

### DIFF
--- a/views/doc.cfm
+++ b/views/doc.cfm
@@ -226,7 +226,6 @@
 			<cfset example_index = 0>
 			<cfloop array="#data.examples#" index="ex">
 				<cfset example_index = example_index + 1>
-				<br />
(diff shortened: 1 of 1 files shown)


pfreitag:
Awesome did you test this one locally?... just want to make sure cuz this type of change can break the entire site if there is an issue. Once I know it has been tested I'll go ahead and merge.

shaedrich:
Now I tested it locally and it worked fine.

PR #964, opened 2018-10-12
Introduce new test

shaedrich:
Check if there's already a doc with the same name as the guide the contributer wants to create

### DIFF
--- a/tests/tests/TestNamesake.cfc
+++ b/tests/tests/TestNamesake.cfc
@@ -1,14 +1,16 @@
 component extends="testbox.system.BaseSpec" {
 	function run(testResults, testBox) {
 		guidesDir = ExpandPath("../guides/en");
-		guides = directoryList(guidesDir, false, "array");
+		guides = directoryList(guidesDir, false, "array").map(function(path) {
+			return getFileFromPath(path);
+		});
 		// temporary: should be removed later
 		index.components = ["application-cfc"];
 
 		describe("JSON Related Tests", function() {
 			it("there should be no guides and function/tag with the same name", function() {
 				for(guide in guides) {
-					expect(fileExists('../data/en/#guide#.json')).toBeTrue('You cannot create with a guide with the name "#guide#" because there is already a doc with the same name.\nTip: Do not hesitate to rename your guide. The title shown in navigation is the title you write into the first line ;-&##41;');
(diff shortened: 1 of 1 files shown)


pfreitag:
So it says it is failing because we have both a doc and a guide named application... but I don't see https://github.com/foundeo/cfdocs/edit/master/data/en/application.json existing?

tests/tests/TestNamesake.cfc
```diff
@@ -0,0 +1,18 @@
...
+		guides = directoryList(guidesDir, false, "array").map(function(path) {
+			return getFileFromPath(path).replace('.md','');
+		});
+		// temporary: should be removed later
+		index.components = ["application-cfc"];
+
+		describe("JSON Related Tests", function() {
+			it("there should be no guides and function/tag with the same name", function() {
+				for(guide in guides) {
+					expect(not (guide is 'application' OR not fileExists('../data/en/#guide#.json'))).toBeTrue('You cannot create with a guide with the name "#guide#" because there is already a doc with the same name.<br />Tip: Do not hesitate to rename your guide. The title shown in navigation is the title you write into the first line ;-)');
```
pfreitag: I'm finding this line confusing because of double negative, you can use toBeFalse() instead of adding the not, and that might help me understand it better... 

Maybe the guide is 'application' part can be moved up into another statement... `if (guide == 'application') { continue; }`

shaedrich: Okay, cool. I'll change that!

shaedrich: Now it claims to have a doc named "cgi-scope" which is not true...

tests/tests/TestNamesake.cfc
```diff
@@ -0,0 +1,21 @@
...
+		// temporary: should be removed later
+		index.components = ["application-cfc"];
+
+		describe("JSON Related Tests", function() {
+			it("there should be no guides and function/tag with the same name", function() {
+				for(guide in guides) {
+					if (guide is 'application') {
+						continue;
+					}
+					expect(fileExists('../data/en/#guide#.json')).toBeTrue('You cannot create with a guide with the name "#guide#" because there is already a doc with the same name.<br />Tip: Do not hesitate to rename your guide. The title shown in navigation is the title you write into the first line ;-)');
```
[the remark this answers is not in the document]
shaedrich: It works! Finally&hellip;^^

pfreitag:
it passes now... merged.

shaedrich:
Sorry for all that unsuccessful commits&hellip; &#42;sigh&#42;

PR #967, opened 2018-10-15
Update isFileObject

Xcreen:
- Added Description
- Added isObject as related
- Changed Parameter (it must be a filestream-object not a path)
- Changed Example

### DIFF
--- a/data/en/isfileobject.json
+++ b/data/en/isfileobject.json
@@ -1,12 +1,12 @@
 {
 	"name":"isFileObject",
 	"type":"function",
-	"syntax":"isFileObject(filePath);",
+	"syntax":"isFileObject(object);",
 	"returns":"boolean",
-	"related":["isPDFObject","isSpreadsheetObject"],
-	"description":"",
+	"related":["isPDFObject","isSpreadsheetObject","isObject"],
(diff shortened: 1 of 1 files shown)


data/en/isfileobject.json
```diff
@@ -1,23 +1,24 @@
...
-	"related":["isPDFObject","isSpreadsheetObject"],
-	"description":"",
+	"related":["isPDFObject","isSpreadsheetObject","isObject"],
+	"description":"Checks if the given object is a filestream-object. A filestream-object you usally get, when calling fileOpen-function.",
 	"params": [
-		{"name":"filePath","description":"The path to the file you are checking.","required":true,"default":"","type":"","values":[]}
+		{"name":"object","description":"The object you want to check, if it is a filestream-object.","required":true,"default":"","type":"","values":[]}
 	],
 	"engines": {
 		"coldfusion": {"minimum_version":"11", "notes":"", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-in-k/IsFileObject.html"}
```
pfreitag: Is this function ACF only? does Lucee also support it?

pfreitag: Looks like lucee does not support it.

PR #968, opened 2018-10-15
Makes externallist an actual array

ianosullivan:
Makes externallist an actual array (not a string) so the code will work when run in the engine.

### DIFF
--- a/data/en/listmap.json
+++ b/data/en/listmap.json
@@ -29,6 +29,6 @@
 	 "examples":[{
 		"title": "Script Syntax",
 		"description": "",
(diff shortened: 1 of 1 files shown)


shaedrich:
Why should a list for a list function be an array?

ianosullivan:
Perhaps I was a bit hasty. I ran the code and saw this error when I ran the code - http://take.ms/VP5AT - and thought is was meant to be an array.

shaedrich:
For some reason the right member function is `str.listAppend()`.

ianosullivan:
Yep list member functions have 'list' as a prefix as per here - [https://cfdocs.org/member](https://cfdocs.org/member) so perhaps that line should be fixed instead.

shaedrich:
I added the member syntax to `listAppend()` in #972. Feel free to make a pull request to fix the example of `listMap()`.

PR #973, opened 2018-10-16
Added result option and example Get Primary Key of INSERTed record

edbartram:
Added result option and example Get Primary Key of INSERTed record


pfreitag:
Thanks!

PR #974, opened 2018-10-17
Update dateDiff.json

coolxdude:
There is difference in dateDiff(datepart, date1, date2) and date1.diff(datepart, date2).
What is actually:
dateDiff(datepart, date1, date2) ==> date2 - date1
date1.diff(datepart, date2) ===> date1 - date2
What was documented here:
dateDiff(datepart, date1, date2) ==> date2 - date1
date1.diff(datepart, date2) ===> date2 - date1

### DIFF
--- a/data/en/datediff.json
+++ b/data/en/datediff.json
@@ -2,7 +2,7 @@
 	"name":"dateDiff",
 	"type":"function",
 	"syntax":"dateDiff(datepart, date1, date2)",
-	"member":"date1.diff(datepart, date2)",
+	"member":"date2.diff(datepart, date1)",
 	"returns":"numeric",
 	"related":["dateadd","dateformat"],
 	"description":"Determines the integer number of datepart units by which date1 is less than date2.",


coolxdude:
this changes is for cold fusion

shaedrich:
This behavior differs in ACF and Lucee:

* [Adobe ColdFusion 2018](https://trycf.com/gist/df4796e9f5252d781b104ebb2da3f595/lucee5?theme=monokai)
* [Lucee 5](https://trycf.com/gist/df4796e9f5252d781b104ebb2da3f595/lucee5?theme=monokai)

data/en/datediff.json
```diff
@@ -2,7 +2,7 @@
 	"name":"dateDiff",
 	"type":"function",
 	"syntax":"dateDiff(datepart, date1, date2)",
-	"member":"date1.diff(datepart, date2)",
+	"member":"date2.diff(datepart, date1)",
```
shaedrich: It's already in there

pfreitag:
Hmm... I think it is probably too late to fix in either engine implementation, that is unfortunate. 

While it's great it is in the engine notes, I wonder if we should have a compatibility warning, that would be more prominent?

shaedrich:
I can't really think of how to make it more prominent than it already is. To do that we might need separate syntaxes per engine.

pfreitag:
I was thinking like a banner when we have a discouraged flag set. For me personally I will not use the `diff` member function because of this incompatibility.

shaedrich:
And what would you like to display compared to what already is in the engine notes?

natewaite:
Could there be a flag that would be strict ACF compatibility in Lucee that would align the behavior with ACF

pfreitag:
Ok I talked to @bdw429s and he suggested that Lucee will change the it to match ACF because it is incompatible if a bug is filed. I have filed a bug here: https://luceeserver.atlassian.net/browse/LDEV-2044 so I'm going to merge this PR because it has the ACF functionality.

PR #980, opened 2018-10-17
Fix errors on getPrinterList()

mjclemente:


data/en/getprinterlist.json
```diff
@@ -17,8 +17,8 @@
 	"examples": [{
 		"title": "Return comma separated printer list",
 		"description": "",
-		"code": "getPrinterList(\", \")",
+		"code": "writeDump(getPrinterList());",
```
shaedrich: Why dumping when it's a string?

mjclemente: So that when the example is run on TryCF the user can see that the it is working. Otherwise, with ACF, no result is show (blank string), which could be confusing.

shaedrich: It's still empty if you dump it. Just with the result, that it's not testable nor does it have a displayable result.

Maybe better mention that the result will be empty because trycf doesn't have any printers or something like that…

mjclemente: Yes, the string is still empty if you dump it, but the result is visible to the user. ACF shows `[empty string]` and Lucee indicates the the result is a string with no length. I checked how Lucee displays their example, and the also use `writedump()`. That said, if you prefer it without, it doesn't make a difference to me.

mjclemente:
Updated example to not use `writedump()`, with added description explaining how the result may be an empty string.

PR #981, opened 2018-10-17
Fix broken license link

mjclemente:


CONTRIBUTING.md
```diff
@@ -27,7 +27,7 @@ Use `true`/`false`
 
 ### How are my contributions licensed?
 
-Please see the [LICENSE](/foundeo/cfdocs/LICENSE) file.
+Please see the [LICENSE](https://github.com/foundeo/cfdocs/blob/master/LICENSE) file.
```
shaedrich: I guess, this might suffice.
```suggestion
Please see the [LICENSE](LICENSE) file.
```

mjclemente: I don't think that works when the rendered docs are presented on cfdocs.org. The result would take you to: https://cfdocs.org/license - which 404s.

mjclemente: I'll handle it however you like.

shaedrich: I guess, the way you did it, is better :+1:

Well, finally it's @pfreitag's decision. I'm just an ordinary contributor^^

pfreitag:
Merged it - that works.

PR #984, opened 2018-10-18
Update listitemtrim.json

Xcreen:
- Added Description
- Added Parameter-Description
- Added Examples

Is there a way to disabled auto-linking in the parameter/example-description?
Because "empy" is auto-linking to the function empty, which is not appropriate.


shaedrich:
Not right now, I guess. Maybe we should open a new issue with that question to discuss that with @pfreitag. Maybe we could prefix the works which would be linked by default but should not in this case.

shaedrich:
@pfreitag What do you think about the question above?

PR #989, opened 2018-10-18
Update cfqueryparam doc, add cfsqltype cheatsheet guide.

### LINKED ISSUE #985: cfqueryparam SQL type mapping table
May not be possible given the JSON structure of all of the pages, but I believe it'd be incredibly helpful for the cfqueryparam page to include the SQL type chart found on https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-p-q/cfqueryparam.html.

AntiTcb:
Resolves #985. 

Preliminarily asking: Does the `values` array allow the CFVersion+ placeholders? I'd like to denote the CF versions that the `cfsqltype` values were added in on the same line that the value is declared.


guides/en/cfsqltype-cheatsheet.md
```diff
@@ -0,0 +1,179 @@
...
+- CF_SQL_VARBINARY
+    - JDBC : VARBINARY
+    - DB2 : Rowid
+    - Oracle : raw
+    - SQL Server : varbinary
+- CF_SQL_VARCHAR
+    - JDBC : VARCHAR
+    - DB2 : Varchar
+    - Oracle : varchar2, nvarchar2
+    - SQL Server : varchar
```
shaedrich: Not very readable but okay until we have table support ;)

data/en/cfqueryparam.json
```diff
@@ -37,39 +38,18 @@
...
         "cf_sql_real",
         "cf_sql_refcursor",
         "cf_sql_smallint",
+        "cf_sql_sqlxml",
         "cf_sql_time",
         "cf_sql_timestamp",
         "cf_sql_tinyint",
         "cf_sql_varchar",
-        "cf_sql_nvarchar",
-        "bigint",
```
shaedrich: It's not a mistake, that they are listed both with "cf_sql_" prefix and without because they work both. But yeah, on the other hand, this is explained in the param and the engine notes. So it's not completely wrong that we don't need both&hellip;

shaedrich:
You want to do something like `{ ..., values: ["cf_sql_bigint", "CF11+ Lucee4.5+ bigint", ... ] }`? I guess, currently it's impossible.

But maybe @pfreitag wants to add this behavior.

guides/en/cfsqltype-cheatsheet.md
AntiTcb: Willing to take other suggestions, but nested lists was the best I could think of for now. Definitely in favor of a table once that's supported.

AntiTcb:
Something more along the lines of `values: [ "cf_sql_nclob CF10+ Lucee4.5+", ... ]` would be ideal for displaying the version alongside the value.

guides/en/cfsqltype-cheatsheet.md
shaedrich: Works for now. I guess, it's the best, if @pfreitag merges this one and you or someone else creates a new pr when table support is added ;-)

shaedrich:
"More along the lines"? Adding the engine version behind is inconsistent.

data/en/cfqueryparam.json
```diff
@@ -37,39 +38,18 @@
...
-        "numeric",
-        "real",
-        "refcursor",
-        "smallint",
-        "time",
-        "timestamp",
-        "tinyint",
-        "varchar",
-        "nvarchar"
+        "cf_sql_nvarchar"
```
KamasamaK: `cf_sql_nvarchar` is in this array twice.

data/en/cfqueryparam.json
pfreitag: I think we want to keep both in there... because editors like Sublime or VSCode will use this info for autocomplete. Can you add them back in and then I'll merge the PR - thanks!

shaedrich: Ah, because of that we left them in.

pfreitag: Yeah also for that reason we don't want to add version support tags in there.

PR #993, opened 2018-10-22
Update array.json

Xcreen:
- Extend description
- Added Examples

### DIFF
--- a/data/en/array.json
+++ b/data/en/array.json
@@ -4,13 +4,28 @@
 	"syntax": "array()",
 	"returns": "array",
 	"related": ["arrayNew"],
-	"description": "Creates a new array populated with any arguments that have been passed to the function.",
+	"description": "Creates a new array populated with any arguments that have been passed to the function. It is recommend by lucee to use the inline array notation.",
 	"params": [
 		{"name":"values","description":"One or more values that can be used in the initialization of the array.","required":false,"default":"","type":"any","values":[]}
 	],
 	"engines": {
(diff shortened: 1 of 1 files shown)


data/en/array.json
```diff
@@ -4,13 +4,28 @@
 	"syntax": "array()",
 	"returns": "array",
 	"related": ["arrayNew"],
-	"description": "Creates a new array populated with any arguments that have been passed to the function.",
+	"description": "Creates a new array populated with any arguments that have been passed to the function. It is recommend by lucee to use the inline array notation.",
```
pfreitag: I see you added `It is recommend by lucee to use the inline array notation.` so are you saying that they discourage using this function? If so you can add a key for this, eg:

`"discouraged": "Lucee recommends using the inline array notation instead of the array function. See example below for details",`

Xcreen:
Thanks for the information :) 
I switched it to the discouraged-key.

data/en/array.json
```diff
@@ -4,7 +4,8 @@
 	"syntax": "array()",
 	"returns": "array",
 	"related": ["arrayNew"],
-	"description": "Creates a new array populated with any arguments that have been passed to the function. It is recommend by lucee to use the inline array notation.",
+	"description": "Creates a new array populated with any arguments that have been passed to the function.",
+	"discouraged": "It is recommend by lucee to use the inline array notation.",
```
shaedrich: Could be simplified to "Lucee recommends to …"

PR #1000, opened 2018-10-24
Added some missing descriptions and other improvements

atuttle:
ref #540

### DIFF
--- a/data/en/queryevery.json
+++ b/data/en/queryevery.json
@@ -5,20 +5,62 @@
 	"member": "query.every(closure [, parallel] [, maxThreads])",
 	"returns": "boolean",
 	"related": [],
-	"description": "",
+	"description": "Executes a callback/closure against every row in a query and returns true if the callback/closure returned true for every row.",
 	"params": [
-		{"name": "query", "description": "", "required": true, "default": "", "type": "query", "values": []},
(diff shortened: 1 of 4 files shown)


data/en/queryevery.json
```diff
@@ -5,20 +5,62 @@
...
+		{
+			"name": "query",
+			"description": "",
+			"required": true,
+			"default": "",
+			"type": "query",
+			"values": []
+		},
+		{
+			"name": "closure",
```
shaedrich: You can use `callback_params` here like so: https://github.com/foundeo/cfdocs/blob/3f33f8d8aacec842205e93d81242cac4d65eac66/data/en/listmap.json#L10

pfreitag: I'll merge this PR, @atuttle if you have a chance please add the `callback_params` in another PR.

atuttle: I can do that... I don't see the point though. How is that data used? The list of callback params displayed on https://cfdocs.org/listmap is just from the description attribute of the function param.

shaedrich: You're right. This is not as handy as possible yet. But maybe this can be improved.

KamasamaK: Having the data is handy in itself. The web site is just one potential consumer for the data, and it can be modified over time to utilize it better if it's available. Another use for the data is in editors to facilitate CFML development.

PR #1003, opened 2018-10-24
Added example for the getDirectoryFromPath function.

notthemonkee:


data/en/getdirectoryfrompath.json
```diff
@@ -14,5 +14,14 @@
...
 	},
-	"links": []
+	"links": [],
+	"examples":[
+		{
+			"title":"Basic Example",
+			"description":"",
+			"code":"directoryPath = getDirectoryFromPath(\"C:\\temp\\file.txt\");",
+			"result":"The directoryPath variable will contain the string C:\\temp\\",
+			"runnable":false
```
pfreitag: Quick thought, if you change the code example to simply this:

getDirectoryFromPath("c:\temp\file.txt")

then set result to c:\temp\ 

This will actually be runnable even on linux, since it is not actually using the file system, and trycf does not block getDirectoryFromPath.

notthemonkee: Got it, thanks @pfreitag I get how this works now.

PR #1004, opened 2018-10-24
Example for getFileFromPath function

notthemonkee:


data/en/getfilefrompath.json
```diff
@@ -14,5 +14,14 @@
 		"railo": {"minimum_version":"", "notes":"", "docs":"http://railodocs.org/index.cfm/function/getfilefrompath"},
 		"openbd": {"minimum_version":"", "notes":"", "docs":"http://openbd.org/manual/?/function/getfilefrompath"}
 	},
-	"links": []
+	"links": [],
+	"examples" : [
+		{
+			"title":"Basic Example",
+			"description":"",
+			"code":"fileName = getFileFromPath(\"C:\\temp\\file.txt\"));",
```
pfreitag: for code you can simply use: `getFileFromPath(\"C:\\temp\\file.txt\"))`

data/en/getfilefrompath.json
```diff
@@ -14,5 +14,14 @@
...
+			"result":"The fileName variable will contain the string file.txt",
```
pfreitag: Just set the result to the actual result, which is: `"file.txt"`

data/en/getfilefrompath.json
```diff
@@ -14,5 +14,14 @@
...
+			"runnable":false
```
pfreitag: If the above changes are made this can be set runnable to `true`

notthemonkee: Got it, thanks @pfreitag I get how this works now.

pfreitag:
Thanks for fixing... that change also allows the Travis-Ci task to validate that the result is correct as well when it runs its tests.

PR #1006, opened 2018-10-24
Add missing descriptions and an example

turqmr2:


data/en/getcurrentcontext.json
```diff
@@ -4,11 +4,19 @@
 	"syntax": "getCurrentContext()",
 	"returns": "array",
 	"related": ["getContextRoot"],
```
pfreitag: add `callstackGet()` as related?

pfreitag: feel free to add that in another PR as I've merged this one.

PR #1008, opened 2018-10-24
datasourceFlushMetacache

shawnoden:
Added description and engine note to datasourceFlushMetacache.


data/en/datasourceflushmetacache.json
```diff
@@ -4,12 +4,12 @@
...
 	"related": [],
-	"description": "Flush the meta data stored for a certain datasource used for stored procedures, this is only supported for Oracle datasources.",
+	"description": "Flush the meta data stored for a certain datasource used for stored procedures.\n NOTE: This is only supported for Oracle datasources.",
 	"params": [
-		{"name": "datasourceName", "description": "", "required": false, "default": "", "type": "string", "values": []}
+		{"name": "datasourceName", "description": "Name of the datasource to flush. When not defined all caches are flushed.", "required": false, "default": "", "type": "string", "values": []}
 	],
 	"engines": {
-		"lucee": {"minimum_version": "", "notes": "", "docs": "http://docs.lucee.org/reference/functions/datasourceflushmetacache.html"}
+		"lucee": {"minimum_version": "4.5", "notes": "This was added in Railo 3.3.", "docs": "http://docs.lucee.org/reference/functions/datasourceflushmetacache.html"}
```
shaedrich: Instead you could add Railo as engine and 3.3 as its `minimum_version`.

shawnoden: @shaedrich I tried adding Railo at first, but I don't think it showed up on my local version. So since there is no version of Lucee below 4.5, I added that as `minimum_version` and added a note that it was added in Railo 3.3. Is Railo still a valid option for "engines"?

shaedrich: It is. And Lucee 4.5 was right :wink:

pfreitag: Yeah it's probably best to add the railo engine to put the minimum version in there... I don't show railo on the cfdocs web page anymore since it is an unsupported/abandoned engine at this point. But someome might find that info handy.

shawnoden: I changed it to add "railo" to the engines with version 3.3, but I didn't include a link for docs, since I'm not really a fan of treating unaffiliated links as "official". If someone associated with Railo owns `it-tallaght.ie` then I will add it. But if not, I don't think we should share a link that could contain whatever that owner wants to include on that page.

PR #1015, opened 2018-10-25
Add related

shaedrich:


data/en/arrayindexexists.json
```diff
@@ -4,7 +4,7 @@
 	"syntax": "arrayIndexExists(array, index)",
 	"member": "array.indexExists(index)",
 	"returns": "boolean",
-	"related": [],
+	"related": ["arrayContains","structKeyExists"],
```
KamasamaK: `arrayIsDefined` is basically the same function, so it probably belongs here too

shaedrich: Oh, thanks. I guess, I already thought of that but its name didn't came to my mind :D

PR #1019, opened 2018-10-25
Created cfupdate example

homestar9:
First example for how you might use cfupdate using cfscript


pfreitag:
👍 thanks for making your first contribution @homestar9

PR #1021, opened 2018-10-26
Add related

shaedrich:


pfreitag:
@shaedrich sorry, it looks like there is a merge conflict with isBinary

shaedrich:
@pfreitag No problem. Resolved it.

PR #1024, opened 2018-10-26
Update systemcacheclear.json

Xcreen:
- Added Parameter Description/Values

### DIFF
--- a/data/en/systemcacheclear.json
+++ b/data/en/systemcacheclear.json
@@ -6,7 +6,22 @@
 	"related": [],
 	"description": "Clears all caches or the specified cache",
 	"params": [
-		{"name": "cacheName", "description": "", "required": false, "default": "", "type": "string", "values": []}
+		{
+			"name": "cacheName", 
+			"description": "Name of the cache you want to clear. If not specified it will clear all caches.", 
+			"required": false, 
+			"default": "", 
+			"type": "string", 
+			"values": [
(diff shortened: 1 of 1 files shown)


data/en/systemcacheclear.json
```diff
@@ -6,11 +6,41 @@
...
 	"params": [
-		{"name": "cacheName", "description": "", "required": false, "default": "", "type": "string", "values": []}
+		{
+			"name": "cacheName", 
+			"description": "Name of the cache you want to clear. If not specified it will clear all caches.", 
+			"required": false, 
+			"default": "", 
+			"type": "string", 
+			"values": [
+				"all (clear all caches)",
```
KamasamaK: `values` should only contain the values themselves, not descriptions.

Xcreen: Okay, but what if a value can contains aliases and means the same?  
E.g.  template or page  means both cleaing the template cache.

shaedrich: Then they both should be included, because IDE plugins can provide both values for autocomplete.

Xcreen: So how you recommend to write this? As seperate values without description? But than the information is lost that page/template means the same.

shaedrich: You can add a note to the param's description.

PR #1034, opened 2018-10-30
Update getnumericdate3

shawnoden:
Corrections to function name based on Lucee docs. RE: Issue foundeo#898.

### DIFF
--- a/data/en/getnumericdate.json
+++ b/data/en/getnumericdate.json
@@ -1,15 +1,16 @@
 {
 	"name": "getNumericDate",
 	"type": "function",
-	"syntax": "getNumericDate(date)",
+	"syntax": "GetNumericDate(arg1)",
 	"returns": "numeric",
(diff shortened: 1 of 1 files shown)


data/en/getnumericdate.json
```diff
@@ -1,15 +1,16 @@
 {
 	"name": "getNumericDate",
 	"type": "function",
-	"syntax": "getNumericDate(date)",
+	"syntax": "GetNumericDate(arg1)",
```
pfreitag: Let's keep this as lowerCamelCase (keep lowercase g)

shawnoden: So we're only changing the Arguments? I do prefer lowerCamelCase, but that casing matched Lucee docs. If that is the preferred way, 'll leave the functions syntax at loweCamelCase.

shawnoden: Done.

PR #1038, opened 2018-11-02
Deprecation notice for imageDrawImage()

shaedrich:


KamasamaK:
What is the meaning of `maximum_version`? We have `deprecated` and `removed` for their respective purposes.

data/en/imagedrawimage.json
```diff
@@ -12,8 +12,8 @@
 		{"name": "y", "description": "", "required": true, "default": "", "type": "numeric", "values": []}
 	],
 	"engines": {
-		"lucee": {"minimum_version": "", "notes": "", "docs": "http://docs.lucee.org/reference/functions/imagedrawimage.html"}
+		"lucee": {"minimum_version": "4.5", "maximum_version": "5", "notes": "This function is deprecated, use ImagePaste instead.", "docs": "http://docs.lucee.org/reference/functions/imagedrawimage.html"}
```
KamasamaK: What is the meaning of `maximum_version`? We have `deprecated` and `removed` for their respective purposes.

pfreitag: Yup should probably be deprecated="5" right?

shaedrich: Mea culpa! Of course!

PR #1039, opened 2018-11-02
example for structEach was using arrayEach

aliaspooryorik:
example for structEach was using arrayEach and caused an error when run.


pfreitag:
Thanks!

PR #1049, opened 2018-11-14
Move text inside ribbon

shaedrich:


shaedrich:
bildschirmfoto 2018-11-13 um 16 02 30
In docs which have old cache data the ribbon markup is old but the css is already up-to-date, which causes that result shown in the picture above.

PR #1057, opened 2018-11-22
Fix and improve callback_params output

shaedrich:


pfreitag:
Just want to make sure this one is tested since it is a view change, please confirm - thanks!

pfreitag:
Sorry closed it by accident when I commented.

pfreitag:
Hmm looking at this the only difference I see is that it switched the h4 to strong? Was this done in another PR if so can we close this one?

shaedrich:
Yupp, that's all

shaedrich:
Thanks!

PR #1063, opened 2018-12-10
Params listed twice (in description as well)

shaedrich:


data/en/cfdirectory.json
```diff
@@ -14,8 +14,8 @@
 		{"name":"sort","description":"Query column(s) by which to sort directory listing.\n Delimited list of columns from query output.","required":false,"default":"asc","type":"string","values":["asc","desc"]},
 		{"name":"newdirectory","description":"New name for directory.","required":false,"default":"","type":"string","values":[]},
 		{"name":"recurse","description":"Whether ColdFusion performs the action on subdirectories.","required":false,"default":false,"type":"boolean","values":[true,false]},
-		{"name":"type","description":"* file: includes only filenames.\n * dir: includes only directory names.\n * all: includes both filenames and directory names.","required":false,"default":"all","type":"string","values":["dir","file","all"]},
-		{"name":"listinfo","description":"* all: includes all information in the result set.\n * name: includes only filenames in the result set.","required":false,"default":"all","type":"string","values":["name","all"]}
+		{"name":"type","description":"","required":false,"default":"all","type":"string","values":["dir","file","all"]},
```
pfreitag: Should we put something in the description or did you think the all, dir, and file are self explanatory?

shaedrich: The description didn't explain more than the values. It's just more text. I would just do it this way when there were special cases for a specific value.

pfreitag: Ok, yeah I agree it wasn't adding anything

PR #1064, opened 2018-12-10
Remove true/false from values

shaedrich:


pfreitag:
@KamasamaK or @jcberquist does this have any issue with auto complete in your ide plugins or can they detect that the type is boolean and provide the suggestions?

KamasamaK:
The VS Code extension can do this, though it only works for tag attributes currently. I also found a bug thanks to this that if the boolean `true` was used instead of the string that it was not working anyway. I have modified it so that it will always present `true` and `false` for booleans in the next version.

shaedrich:
Good point :+1:

shaedrich:
@pfreitag 
> @jcberquist **via Slack**: I don't think this change will be a problem for the ST plugin I maintain since I actually compile a list of completions ahead of time into JSON files. So it shouldn't be affected directly by this, but I am glad to know for the sake of the next time I update the JSON files.

pfreitag:
Ok sounds like this should be good to merge then correct?

shaedrich:
That's how it sounded to me as well.

PR #1068, opened 2018-12-20
Fix invalid currentrow usage

nbokoski:
`currentrow` is a member of the query object, not the index.


pfreitag:
Nice catch -- thanks for contributing!

PR #1074, opened 2019-01-04
Make comparison consistent with the rest of the document

savaticus:
The document should be consistent, it was not and I decided to change it.


pfreitag:
Thanks @savaticus

shaedrich:
@savaticus @pfreitag Maybe you should add the engines and versions as stated in the compatibility notes.

PR #1075, opened 2019-01-04
listAppend returns the concatenated list

savaticus:
This doc is a bit confusing when you are new to CF because unlike every other CF append function (array, struct) it isn't atomic, even when the member syntax is used. I added this to the description and the code examples to highlight this. listPrepend's doc clearly shows the proper use. Which is how I figured out what was wrong in my own code.

### DIFF
--- a/data/en/listappend.json
+++ b/data/en/listappend.json
@@ -28,7 +28,7 @@
 			{
 				"title":"Simple listAppend Example with Delimiter",
 				"description":"Add 'foo' to the end of this list using a custom delimiter",
-				"code" : "newList = listAppend(\"bar|bar2\", \"foo\", \"|\")",
+				"code" : "oldList = \"bar,bar2\";\nnewList = listAppend(oldList, \"foo\");\nwriteOutput(oldList & \"-->\" & newList);",
 				"result":"bar|bar2|foo"
 		}
 	]
--- a/data/en/listappend.json
+++ b/data/en/listappend.json
@@ -22,7 +22,7 @@
 		{
 				"title":"Simple listAppend Example",
 				"description":"Add 'foo' to the end of this list",
-				"code" : "newList = listAppend(\"bar,bar2\", \"foo\")",
+				"code" : "oldList = \"bar,bar2\";\nnewList = listAppend(oldList, \"foo\");\nwriteOutput(oldList & \"-->\" & newList);",
 				"result":"bar,bar2,foo"
 		},
 			{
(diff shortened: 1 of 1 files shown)


shaedrich:
It doesn't modify the original string, because it is a simple value which is not held by reference in the variable.

data/en/listappend.json
```diff
@@ -22,13 +22,13 @@
 		{
 				"title":"Simple listAppend Example",
 				"description":"Add 'foo' to the end of this list",
-				"code" : "listAppend(\"bar,bar2\", \"foo\")",
+				"code" : "newList = listAppend(\"bar,bar2\", \"foo\")",
```
shaedrich: Maybe this is what you attempted to emphasize:

```suggestion
				"code" : "oldList = \"bar,bar2\";\nnewList = listAppend(oldList, \"foo\");\nwriteOutput(oldList & \"-->\" & newList);",
```

data/en/listappend.json
```diff
@@ -22,13 +22,13 @@
...
 				"result":"bar,bar2,foo"
 		},
 			{
 				"title":"Simple listAppend Example with Delimiter",
 				"description":"Add 'foo' to the end of this list using a custom delimiter",
-				"code" : "listAppend(\"bar|bar2\", \"foo\", \"|\")",
+				"code" : "newList = listAppend(\"bar|bar2\", \"foo\", \"|\")",
```
shaedrich:
```suggestion
				"code" : "oldList = \"bar,bar2\";\nnewList = listAppend(oldList, \"foo\");\nwriteOutput(oldList & \"-->\" & newList);",
```

data/en/listappend.json
savaticus: I think that would emphasize it, but if you look at the doc for listPrepend you will see that the examples are similar to my edit here. I was merely attempting to make it easier for others to see that unlike arrayAppend and structAppend, listAppend does not modify the list specified but instead returns a copy with the new value appended.

pfreitag:
FYI - travis is throwing an exception when it is running the example:

`Failure: listappend.json example result is:bar,bar2,foo but evaluated to:EXCEPTION in example 1 in listappend.json: missing required argument [string] for function [writeoutput]. Expected [bar,bar2,foo] Actual [EXCEPTION in example 1 in listappend.json: missing required argument [string] for function [writeoutput]]`

shaedrich:
@pfreitag This is what I wanted to cover with my suggestion above.

data/en/listappend.json
shaedrich: Your example didn't have a `writeOutput()` so it caused the tests to fail like Pete mentioned below.

PR #1078, opened 2019-01-13
Initial schema changes for new member_details

KamasamaK:
As discussed in #525, here are the proposed changes. I would still be interested in feedback from @pfreitag  regarding my latest comments in that issue.

### DIFF
--- a/data/cfdocs.schema.json
+++ b/data/cfdocs.schema.json
@@ -168,7 +168,7 @@
   "properties": {
     "name": {
       "title": "Name",
-      "description": "The name of the tag or function. Use lowercase.",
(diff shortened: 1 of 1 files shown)


pfreitag:
@KamasamaK - I can think of one case where the first argument is not the object being operated on... `find` and `reFind`

pfreitag:
Otherwise it looks good... maybe we just have to have a index variable that tells which param it operates on (defaulted to first, 1) so we could still avoid duplicating params.

KamasamaK:
Ok, I added a property for the index.

pfreitag:
Thanks for your work on this @KamasamaK

PR #1085, opened 2019-01-30
add index to list description

zspitzer:


KamasamaK:
If you're going to specify what other attributes can be used with it, then it should also mention that `item` can be used as well in Railo/Lucee and CF2016+.

PR #1088, opened 2019-02-09
ArrayReverse example using different syntax in ACF2018

aliaspooryorik:
Not quite sure how to document this, but raising a PR. 

ArrayReverse does not exist in ColdFusion but you can get the same result by using the array slice syntax `[::-1]` adding in ACF 2018. Have added an example  of doing it in ACF2018, but not sure how you'd normally handle this as it's an equivalent.


shaedrich:
That's how we handle this in other docs as well, I guess. So, looks good :+1:

KamasamaK:
This kind of thing was brought up in #812. but there's currently no official way to do it.

PR #1090, opened 2019-02-11
Correct return value on charsetEncode

optik-aper:


pfreitag:
Good catch thanks @optik-aper

optik-aper:
@pfreitag No problem.  Thanks for maintaining this helpful resource.

PR #1092, opened 2019-02-19
Add note about using member function on literals

shaedrich:


pfreitag:
Thanks!

shaedrich:
As always: My pleasure ;-)

PR #1103, opened 2019-03-15
Remove incorrect `action` arg description in cffeed

michaelborn:
This _"The name to use for the JavaScript proxy class."_ doesn't belong - it seems to have come from https://cfdocs.org/cfajaxproxy.


pfreitag:
Thanks!

michaelborn:
Welcome!

PR #1108, opened 2019-03-22
Make description consistent to consistent to `transactionRollback()`'…

shaedrich:
…s `savepoint' param


pfreitag:
Thanks I tried updating the commit on that previous PR with your suggestion, but it didn't take.

shaedrich:
Oh, okay. Did you see an error message or like? Maybe only @michaelborn was allowed to…

Did you reported it via github's feature feedback functionality?

pfreitag:
Yes I think it said I didn't have permission, but I was able to edit other PR commits, just not the suggestion you did.

shaedrich:
Hmmm…Maybe it was a little too short-sighted. But you can report the case you had if you want to: https://www.research.net/r/SuggestedChanges

PR #1111, opened 2019-03-23
Script syntax is supported by Lucee

shaedrich:


shaedrich:
@pfreitag Travis got this:
> xmlvalidate.json example result is:true but evaluated to:EXCEPTION in example 1 in xmlvalidate.json: Read timed out

Maybe it just have to be re-run, right?

KamasamaK:
To be clear, _that_ script syntax is not supported by ACF from what I can tell but you can still use the standard `cfmodule(name="cached")` syntax.

PR #1125, opened 2019-05-23
 Add terminateOnTimeout attr for cfexecute

michaelborn:
Document the terminateOnTimeout=false|true attribute. Also give another
example of how to use `cfexecute()`.


michaelborn:
And sorry for the junky merge commit - if I tried hard enough I could figure out how to squash that / edit that out.

pfreitag:
No prob. thanks!

michaelborn:
:)

PR #1127, opened 2019-05-25
Mark encodeFor* functions as *not* deprecated

michaelborn:
The decision to deprecate these `encodeForHTML()`, `encodeFor*()` functions has been reversed per Brad Wood and others more informed than I. :)
See https://groups.google.com/forum/#!topic/lucee/90xgx_wnVs4,
https://cfml.slack.com/messages/C06TA0A9W/convo/C06TA0A9W-1558749781.016600/

### DIFF
--- a/data/en/encodefor.json
+++ b/data/en/encodefor.json
@@ -3,7 +3,7 @@
 	"type":"function",
 	"syntax":"encodeFor(type, value)",
 	"returns":"string",
-	"related":["encodeForHTML", "canonicalize"],
+	"related":["encodeForHTML", "canonicalize","esapiEncode"],
 	"description":"Encodes a given string for safe output in the specified context. The encoding is meant to mitigate Cross Site Scripting (XSS) attacks.",
 	"params": [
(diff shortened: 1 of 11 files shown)


shaedrich:
This differs between Lucee and ACF:
> They're deprecated in Lucee but they will never go away because of compatibility requirements w/ACF. The behavior is a bit different between the two but you're fine using encodeForHTML.

Do you have an **official** source?

michaelborn:
Does this count as an official source?

https://dev.lucee.org/t/htmleditformat-or-encodeforhtml-or-esapiencode-html/982

Edit: Of course not, but I'm wondering where this conversation took place.

shaedrich:
"Technical Advisory Group" sounds official...

pfreitag:
Hey Folks -- I would indeed avoid calling encodeForHTML, etc deprecated on Lucee, it is an ACF functionality that is at the core of how we should be coding, so I don't think it makes sense to have two conflicting best practices for each engine. 

Thanks for fixing this @michaelborn

PR #1128, opened 2019-06-06
Better example

tomchiverton:
render() doesn't return the output, it outputs there and then. Also the previous example didn't match the "expected" anyway.

### DIFF
--- a/data/en/render.json
+++ b/data/en/render.json
@@ -19,8 +19,8 @@
         {
 		"title": "Execute replace using render",
 		"description": "Description of the code example",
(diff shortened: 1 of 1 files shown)


pfreitag:
Hi @tomchiverton  - thanks for the PR. Travis CI is telling us that the JSON is not valid, can you check on that and update it?

KamasamaK:
The issue is that the double quotes are no longer being escaped.

jordanclark:
I fixed the json in @tomchiverton's pull request, sorry I didn't know how to post the change here.
https://github.com/foundeo/cfdocs/pull/1132

tomchiverton:
Ta. I can't find anyway to close this PR

PR #1131, opened 2019-06-13
Update fileAppend for new ColdFusion updates

KamasamaK:
Added new parameter `addNewLine`, added docs link, and updated engine description based on ColdFusion 2016 Update 11 and ColdFusion 2018 Update 4.


pfreitag:
Thanks!

PR #1132, opened 2019-06-13
Add missing lucee tags and correct lucee doc links

jordanclark:
add missing lucee tags: cfpageencoding, cfforward, cfhtmlbody, cfstopwatch
change all docs.lucee.org links to https which is the canonical url


shaedrich:
According to your overview of changes, they sound good. But putting them all in one pr makes it hard to check them.

jordanclark:
@shaedrich sorry about that, as I've been committing more changes they've just been rolling into the same outstanding pull request. I can't create multiple pull requests unless I create a new branch for each commit. If you want to review them separately can't you just look at the changes in each commit?

shaedrich:
Well, I'm not the one to decide that, but yeah, I highly suggest grouping related changes into the same branch so that you end up with separate pull requests for each specific change with can consist of one or a few more commits. Having 754 different files with different kind of changes — there's no easy way to check that anymore… It's a mess, sorry.

jordanclark:
Okey dokey

Issue #898, opened 2018-08-10
Ensure parameter names match ColdFusion 2018+


KamasamaK:
This refers to the parameter names used in [NamedParametersColdFusion2018.pdf](http://cfdownload.adobe.com/pub/adobe/coldfusion/2018/publicBeta/NamedParametersColdFusion2018.pdf)

I think these are mostly good, but this issue will be a reminder for me (or anyone else) to go through all of them to ensure that they match. Lucee already supported this feature, and they have already ensured via [LDEV-1806](https://luceeserver.atlassian.net/browse/LDEV-1806) that they at least have aliases for the ColdFusion names. Because of this, we should go with the more restrictive name.

KamasamaK:
I've started this and I've found there to be more discrepancies than I expected. Also, there are discrepancies between the PDF and the web docs, so I'm siding with the PDF for now until I can get confirmation. Also, multiple signatures (#529) is complicating it, so for now I'm keeping with combining the params e.g. `value or callback`.

KamasamaK:
@shawnoden  You are correct that I've only done up to C. I've been meaning to get back to this, but I have not done any after C that are pending. A few months ago, I asked Saurav -- the person who works on ColdFusion documentation -- and was told that the PDF is correct and that the web pages would be updated to match. Unfortunately, it doesn't seem like this is being done in a timely manner.

shawnoden:
OK, as I get time, I'll work on these. I'll post the letter I'm working on here both when I start and when I finish so we don't step on each other.

shawnoden:
For `syntax`, if a function has an optional parameter, should we signify that in both the `syntax` block and in the `params`  block? ie `{...."syntax":"myFunction(required1,[optional1])"`.... `params":[ "name":"optional1",......,"required":false, ....]}` or `{...,"syntax":"myFunction(required1,optional1)"...}` << So put in square brackets or not?

KamasamaK:
The square brackets seems to be the preferred way to denote this, but the `required` value in `params` is most important. We could even dynamically generate the syntax string with sufficient information in `params`.

shawnoden:
[...]

PR #1139, opened 2019-06-17
lucee parallel functions

jordanclark:
struct, array & query cleanup & standardize map(), reduce(), filter(), each(), some()
- add missing "parallel" and "maxThreads" arguments
- converted "closure" to function with callback_params


pfreitag:
Hi @jordanclark 

Thanks for this PR, though I don't think I can merge this with multiple syntaxes, we had discussed coming up with a way to show multiple syntaxes a while back - I think @shaedrich might have started investigating coming up with a syntax in the json to support it.

What you could do is keep syntax as is for now, and I can merge the arguments docs at least for now.

jordanclark:
Hi @pfreitag 

I'm not sure I understand the issue that it can't have multiple syntaxes? I'm guessing you mean that it includes lucee only arguments?

The thing is, I've just taken what already exists on cfdocs and standardized it for these tags, if you take a look at https://cfdocs.org/arrayeach you'll see it already has ```[, parallel] & [, maxthreads]```

But then https://cfdocs.org/arrayfilter does not, even though they are similar, so I made them consistent.

My thought was since the arguments are documented as lucee 4.5+ only it would be ok?
If not, what about syntax like ```[(lucee), parallel] [(lucee), maxThreads]```

If not let me know and I can take them out, but I also added the callback functions params to the syntax which I think is useful so I'd still like to keep that at least.

pfreitag:
Hi @jordanclark sorry I think I misread something when reviewing it previously.

KamasamaK:
@jordanclark Please keep in mind that changing the param names should comply with #898. For example, in `arrayEach` you changed `callback` to `function` which is not the name ACF 2018+ expects but is an acceptable alias in Lucee.

jordanclark:
@KamasamaK Sorry I didn't realize ACF had a different param name, I'll correct in another pull request

PR #1144, opened 2019-06-19
Typo errors

sarwanKumar1993:


data/en/dotnettocftype.json
```diff
@@ -6,7 +6,7 @@
 	"related":[],
 	"description":"Explicitly converts a value returned by a .NET method to the corresponding ColdFusion data type.",
 	"params": [
-		{"name":"variableName","description":"Name of the .NET variable to convert.","required":true,"default":"","type":"Varible","values":[]}
+		{"name":"variableName","description":"Name of the .NET variable to convert.","required":true,"default":"","type":"Variable","values":[]}
```
pfreitag: Hmm... should the type actually be "variableName", that is a valid CFML type?

KamasamaK: Given the example at https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-c-d/dotnettocftype.html I think `any` would be more appropriate since it's passing the .NET object and not the name of a variable.

sarwanKumar1993: Yes you are correct @KamasamaK  , type **'any'** is more suitable in this case.

pfreitag:
@sarwanKumar1993 thanks for the PR and for updating it with suggestions. @KamasamaK thanks for your input here!

PR #1147, opened 2019-06-20
Fix message when only one engine supports function/tag

### LINKED ISSUE #1146: "This tag/function requires" - engine not supported
# Current situation
When we display the message mentioned in the title but the engine is not supported, the sentence ends after "requires" without mentioning the required engine.

It seems like we deprecated the other engines so that the ifs only result in supported engines, but we weren't consistent enough to remove the message itself.

https://github.com/foundeo/cfdocs/blob/eb4b57a1b1980f14cdaa35d1c5765959fad23d46/views/doc.cfm#L70-L80

# Examples
- [structListNew](https://cfdocs.org/structlistnew)
- [structValueArray](https://cfdocs.org/structvaluearray)

# Which solution do we choose?
- Removing the message for other engines
- Extend the ifs to cover the other engines as well

Solution 1 seems odd because the link to the docs appears anyway.

# Additional questions
- Why are even ACF and Lucee inconsistent?
    - While ACF lists the exact version number whereas Lucee does not
    - While ACF says "Not supported on Lucee, OpenBD, etc.", Lucee only says "Not supported on Adobe ColdFusion." and that's it

shaedrich:
Fix #1146

### DIFF
--- a/views/doc.cfm
+++ b/views/doc.cfm
@@ -70,12 +70,17 @@
 			<cfelseif StructKeyExists(data, "engines") AND structCount(data.engines) EQ 1>
 				<div class="alert alert-warning">
 					This <cfif data.type IS "tag">tag<cfelseif data.type IS "function">function</cfif> requires
-					<cfif structKeyExists(data.engines, "coldfusion")>
-						Adobe ColdFusion<cfif StructKeyExists(data.engines.coldfusion, "minimum_version") AND Len(data.engines.coldfusion.minimum_version)> #data.engines.coldfusion.minimum_version# and up</cfif>.
-						<em> Not supported on Lucee, OpenBD, etc.</em>
-					<cfelseif structKeyExists(data.engines, "lucee")>
-						Lucee. <em>Not supported on Adobe ColdFusion.</em>
-					</cfif>
+					<cfscript>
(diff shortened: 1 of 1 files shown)


pfreitag:
Looks good upon glancing it over, have you tested this one locally? -- just want to make sure before merging cuz it will cause entire site to break if syntax error.

Probably should add some tests to travis to check that pages load ok

shaedrich:
It didn't check it locally yet, but I checked that codesnippet via trycf because I wasn't at home yet. If you want me to, I can test it locally in about an hour.

So you want a travis test for the pages itself?

pfreitag:
Ok - trycf is close enough... yeah I was thinking we could actually do this in testbox, just hit a few (a tag, a function,a guide) url's with cfhttp and make sure the status code is 200

shaedrich:
Okay, fine! Sounds simple enough :)

pfreitag:
Just committed one: https://github.com/foundeo/cfdocs/blob/master/tests/tests/TestViews.cfc

pfreitag:
Thanks - I re-ran this PR through travis - it passed with the new check in place.

shaedrich:
Wonderful! My pleasure :)

PR #1151, opened 2019-06-25
Fix description for step parameter of cfloop

am2605:


pfreitag:
Good one @am2605 thanks!

PR #1152, opened 2019-06-26
 change param "function" to "callback" for ACF compatibility

jordanclark:
#898 @KamasamaK

### DIFF
--- a/data/en/arrayeach.json
+++ b/data/en/arrayeach.json
@@ -1,8 +1,8 @@
 {
 	"name":"arrayEach",
 	"type":"function",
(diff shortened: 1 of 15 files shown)


pfreitag:
@jordanclark I think this change should only be made on the param names, not the `syntax`, since `function` is a valid keyword there. Let me know what ya'll think.

jordanclark:
Oh ya that makes sense, I'll fix

jordanclark:
@pfreitag corrected, thanks for reviewing

pfreitag:
You're welcome @jordanclark and thank you for the PR!

PR #1153, opened 2019-06-26
ACF "struct" named param compatibility

jordanclark:
More #898 goodness

### DIFF
--- a/data/en/arrayeach.json
+++ b/data/en/arrayeach.json
@@ -1,14 +1,14 @@
 {
 	"name":"arrayEach",
 	"type":"function",
-	"syntax":"arrayEach(array, function(item, [index, [array]]){} [, parallel] [, maxThreads])",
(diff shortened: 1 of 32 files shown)


pfreitag:
@jordanclark can you make the same changes as you did in #1152 here. Thanks

pfreitag:
BTW - thanks for doing these in smaller PR chunks, it is easier for me to review lots of small PRs than one big one.

jordanclark:
@pfreitag Will do. Doing separate PR & branches is good for me, helps with my git foo

KamasamaK:
One thing to ensure in the future is that you're always branching from `master`. This was branched from `callback-params` that you used in #1152, which is why is contains a lot of the same changes from that. See [your network graph](https://github.com/jordanclark/cfdocs/network) to visualize this.

PR #1157, opened 2019-07-04
Improve CFSchedule documentation with CF-only and Lucee-only attributes

michaelborn:
I added the following attributes from CF10:

- onComplete
- eventHandler
- onException
- cronTime
- repeat
- priority
- exclude
- onMisfire
- cluster
- overwrite
- retryCount 

I added the following attributes from Lucee (I assume 4.5+):

*  unique
* readonly
* autodelete

I also added two small examples of CF-only attributes, and added the CF10+ `action` values.


michaelborn:
Feel free, anyone, to poke holes in my documentation-copying tendencies. Note that the CF documentation also notes `isDaily` was added in CF10, but neither describes it nor mentions it in the table of values, so I skipped it for now.

https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-r-s/cfschedule.html

pfreitag:
Thanks @michaelborn

PR #1160, opened 2019-07-17
Add default value for `scale` argument of cfqueryparam

hawkrives:
As shown in https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-p-q/cfqueryparam.html, and as experienced by my upgrade from CF2016 to CF2018 this morning.


pfreitag:
Thanks @hawkrives

PR #1163, opened 2019-07-22
Test more code

turqmr2:
Current code limits testing to one-liners. This enables testing of all code with "runnable:true" and fixes some errors in the example code.


data/en/arrayclear.json
```diff
@@ -23,19 +23,22 @@
 			"title": "Clear the value of an array",
 			"description": "Uses the arrayClear function to clear the value of an array",
 			"code": "someArray = [\"Red\", \"White\", \"Green\", \"Blue\", \"Pink\"];\nwriteOutput(arrayClear(someArray));",
-			"result": "Yes"
+			"result": "Yes",
```
shaedrich: I'm not sure, if it really returns "Yes" here whereas it returns "true" for the next example … oO

turqmr2: Either way, it resolves as true. Though, I agree with you that "true" is preferable.

PR #1164, opened 2019-07-24
2nd try... This should pass travis

turqmr2:
Current code limits testing to one-liners. This enables testing of all code with "runnable:true" and fixes some errors in the example code.


pfreitag:
@turqmr2 just FYI it is easer to review pull requests that don't change so many files, so if you want to resubmit parts of this just do it is smaller chunks and it will get reviewed faster.

PR #1166, opened 2019-08-10
Add Lucee `encodeFor` support note to writeOutput

michaelborn:
Lucee 5.1.0.8+ supports the `encodeFor` attribute on `writeOutput()`. I'm not certain that tag `Lucee5.1.0.8+` will work, but here goes. :)

(If that tag doesn't work, should we update the engine regex or just note it as "Since Lucee 5.2+`?

https://docs.lucee.org/reference/functions/writeoutput.html


pfreitag:
Yeah, I think it will label it properly as long as it is dotted numeric version... merging we'll find out.

Issue #652, opened 2017-10-24
[MAJOR] Markdown can't parse tables?


shaedrich:
When I visit [DateTime Cheatsheet guide](https://cfdocs.org/datetime) the tables are broken. Is [the parser](/rjeschke/txtmark) unable to parse tables?

shaedrich:
@pfreitag Would be nice, if I could go on with this...

pfreitag:
It cannot parse tables. Not sure which is easier:

1) switching parsers 
2) fix that guide to not use tables.

shaedrich:
Tables are essential to simply explain things and very basic markup so I would prefer to switch parsers.

pfreitag:
Ok - I'd like to give this a whirl: https://github.com/vsch/flexmark-java because it is what is used by the ColdBox markdown module (cbmarkdown) and I'd eventually like to switch the site to use ColdBox.

shaedrich:
Sorry for my late reply.
Flexmark seems to have a fair bit of features. Most of these won't probably not used by cfdocs. But to have tables parsed will be a big enhancement :+1:

shaedrich:
Should we do this?

jmharrin:
I just created a Pull Request to address this.

mjclemente:
Just stumbled across the [Datetime Cheatsheet](https://cfdocs.org/datetime) being broken because of the markdown/table issue. Was gonna put in a PR, until I realize the cause was the parser.

FWIW, Flexmark does seem to be more actively maintained.

PR #1172, opened 2019-08-24
Turn CFSqlType_Cheatsheet into a table

hawkrives:
I don't know if the website frontend can handle tables yet; do you need help with that?

### DIFF
--- a/guides/en/cfsqltype-cheatsheet.md
+++ b/guides/en/cfsqltype-cheatsheet.md
@@ -1,179 +1,44 @@
 # CFSqlType Cheatsheet
 
-[Table sourced from Adobe Documentation](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-p-q/cfqueryparam.html)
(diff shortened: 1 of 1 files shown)


KamasamaK:
Based on how the [Date Time Cheatsheet](https://cfdocs.org/datetime) looks, it unfortunately does not support Markdown tables right now.

hawkrives:
It looks like cfdocs uses https://github.com/rjeschke/txtmark.

Perhaps I will find the time to look into a replacement (maybe https://github.com/vsch/flexmark-java, which has … many options)

KamasamaK:
OK. That's the same project @pfreitag proposed in #652 so sounds good.

jmharrin:
I created a pull request yesterday to address this.  I replaced txtmark with an optimized build of flexmark so table markdown will be supported. #1361

KamasamaK:
With #1477 merged, this can probably be revisited.

pfreitag:
Thanks all!

PR #1177, opened 2019-09-07
listFind updates

nick1701d:
Add member function details to listFind. Improve formatting and member function example description in listFindNoCase


pfreitag:
Thanks @nick1701d

PR #1180, opened 2019-09-14
Added CF2016's invokeImplicitAccessor setting

michaelborn:
This was added way back in 2017, so I thought we should document it. :)


pfreitag:
Thanks @michaelborn

PR #1181, opened 2019-09-24
Lucee isValid() update

michaelborn:
Added a bunch of types that Lucee supports in `isValid()`.

https://github.com/lucee/Lucee/blob/dbb220d72027fea68f927654388456515289867a/core/src/main/java/lucee/runtime/op/Decision.java#L972-L1047

Some of them are new and useful, like `isValid("lambda", x)` or `isValid("image", xyz)`, while others are merely aliases. (For example, I love the idea of using `isValid("class", myObject)` over `isValid("component", myObject)`.)

As I type up this PR, I realize that I should have added `Lucee5+` to each Lucee-only type, as I'm certain these are all or mostly all Lucee-only.

Sigh. Maybe tomorrow.

PR #1182, opened 2019-09-24
Ignore case for self-relation test

shaedrich:
Occurred in https://github.com/foundeo/cfdocs/pull/1181#discussion_r327451477


pfreitag:
Thanks @shaedrich

shaedrich:
You're welcome @pfreitag

Issue #541, opened 2017-09-28
Several Tags / Functions do not have examples


pfreitag:
See list here: https://cfdocs.org/reports/missing-examples.cfm

bl2i4n:
Hey @pfreitag, I would like to contribute to this.

pfreitag:
@bl2i4n great -- some good ones to start with `isStruct` `isNull` `isJSON`

Just go here: https://cfdocs.org/isstruct then click on the fork me on github banner on the top right, it would bring you here: https://github.com/foundeo/cfdocs/edit/master/data/en/isstruct.json

From there you can click edit and add an example. Here's an example of a function that has examples: https://github.com/foundeo/cfdocs/edit/master/data/en/hash.json 

You can use this utility: https://cfdocs.org/utilities/json/ to put your example into valid JSON.

zspitzer:
Could a link be added to pre-populate https://cfdocs.org/utilities/json/ with the existing values for an entry?

shaedrich:
The fields map to one single json key-value-pair each. Where should something be pre-filled?

PR #1189, opened 2019-10-03
#541 Examples for arrayGetMetadata

CraicOverflow89:
Simple example and member function. Relates to issue  #541.


pfreitag:
Thanks @CraicOverflow89 appreciate you taking the time to do the pull requests. I can merge your other PRs with a minor fix on the `results`.

CraicOverflow89:
No problem @pfreitag, those should all be sorted now. Made the mistake of thinking the result string was only read by humans.

PR #1191, opened 2019-10-03
#541 Examples for echo

CraicOverflow89:
Simple example and complex data with note to clarify that toString is not called on parameters.

### DIFF
--- a/data/en/echo.json
+++ b/data/en/echo.json
@@ -16,13 +16,15 @@
 			"title": "Simple Example",
 			"description": "Prints a simple string.",
 			"code": "echo(\"hello world\");",
-			"result": "hello world"
+			"result": "hello world",
+			"runnable":true
 		},
 		{
 			"title": "Complex Data Example",
 			"description": "Note that complex objects will not automatically be converted to String, as with System.out.println in Java.",
 			"code": "// Create data\r\ndata = {\r\n    \"title\"  : \"A Game of Thrones\",\r\n    \"author\" : \"George R. R. Martin\",\r\n    \"ISBN\"   : \"0-00-224584-1\"\r\n};\r\n\r\n// Print data\r\necho(data.toString());",
-			"result": "{ISBN={0-00-224584-1}, author={George R. R. Martin}, title={A Game of Thrones}}"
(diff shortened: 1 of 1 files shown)


data/en/echo.json
```diff
@@ -11,6 +11,19 @@
 	"engines": {
 		"lucee": {"minimum_version": "", "notes": "", "docs": "https://docs.lucee.org/reference/functions/echo.html"}
 	},
-	"examples": [],
+	"examples": [
+		{
+			"title": "Simple Example",
+			"description": "Prints a simple string.",
+			"code": "echo(\"hello world\");",
+			"result": "Prints the string 'hello world'"
```
pfreitag: @CraicOverflow89 the result should be the actual result, of the code, so in this case it should just be `"hello world"` instead of `"Prints the string 'hello world'"`

The reason for this is that in some cases the result is actually tested automatically by our continuous integration scripts.

data/en/echo.json
```diff
@@ -11,6 +11,19 @@
...
+		},
+		{
+			"title": "Complex Data Example",
+			"description": "Note that complex objects will not automatically be converted to String, as with System.out.println in Java.",
+			"code": "// Create data\r\ndata = {\r\n    \"title\"  : \"A Game of Thrones\",\r\n    \"author\" : \"George R. R. Martin\",\r\n    \"ISBN\"   : \"0-00-224584-1\"\r\n};\r\n\r\n// Print data\r\necho(data.toString());",
+			"result": "Prints '{ISBN={0-00-224584-1}, author={George R. R. Martin}, title={A Game of Thrones}}'"
```
pfreitag: Same issue here. Please remove `Prints '` and just put the actual value.

KamasamaK:
These should probably set `runnable` to `true`. Then they could be run from the page and the CI can verify the result. (One of them is still incorrect)

CraicOverflow89:
@KamasamaK where am I setting runnable to true?

KamasamaK:
@CraicOverflow89 It should be set in the example object. Check out the [sample JSON](https://github.com/foundeo/cfdocs#json-file-documentation).

CraicOverflow89:
Ok thanks @KamasamaK.

pfreitag:
Thanks @CraicOverflow89

PR #1192, opened 2019-10-03
#541 Examples for getapplicationmetadata

CraicOverflow89:
Added simple example using some of the properties found in ACF and Lucee application meta data.

### DIFF
--- a/data/en/getapplicationmetadata.json
+++ b/data/en/getapplicationmetadata.json
@@ -12,8 +12,14 @@
 		"lucee": {"minimum_version":"", "notes":"", "docs":"https://docs.lucee.org/reference/functions/getapplicationmetadata.html"},
 		"openbd": {"minimum_version":"", "notes":"", "docs":"http://openbd.org/manual/?/function/getapplicationmetadata"}
 	},
-	"links": [
-
-	]
+	"examples" : [
+		{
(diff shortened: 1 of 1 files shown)


data/en/getapplicationmetadata.json
```diff
@@ -12,8 +12,14 @@
...
 	},
-	"links": [
-
-	]
+	"examples" : [
+		{
+			"title": "Simple Example",
+			"description": "",
+			"code": "// Fetch application meta data\r\ndata = getApplicationMetadata();\r\n\r\n// Print application name\r\nwriteOutput(\"Application name is \" & (data.name.length() ? data.name : \"unspecified\") & \"<br>\");\r\n\r\n// Print session timeout\r\nwriteOutput(\"Session timeout is \" & data.sessionTimeout & \"<br>\");\r\n\r\n// Print session management\r\nwriteOutput(\"Session management is \" & (data.sessionManagement ? \"enabled\" : \"disabled\"));",
+			"result": "Prints the above statements using application meta data."
```
pfreitag: The `result` string should be used for the actual computed result, so in this case I would just leave the result empty, and move `Prints the above statements using application meta data.` to the `description`.

pfreitag:
Thanks @CraicOverflow89

PR #1195, opened 2019-10-04
Fixes #1194 by checking length of string instead of struct

### LINKED ISSUE #1194: Error in vew/doc.cfm due to use of Len() on struct when running in ACF
Found this when looking at doc for `lsWeek()` after spinning up the project on CommandBox which defaulted to latest ACF engine. Happens for some other function pages as well.

Appears due to `Len(data.engines[engine])` at `\views\doc.cfm:82`. Lucee has no issue with this and returning a count of keys (not what we want in this case), but ACF complains.

PR coming shortly.

CreativeNotice:


CreativeNotice:
@pfreitag I'm not super familiar with the CI but I think the build error is due to an issue pulling a dependency down from ForgeBox? Is there a way to restart the build?

https://travis-ci.org/foundeo/cfdocs/jobs/593392572

pfreitag:
Yes, I can - I just restarted the build. Thanks for letting me know that's why it failed, saved me a few minutes of looking at the build logs.

PR #1196, opened 2019-10-04
lsWeek() expand local param documentation and add an example.

CreativeNotice:


data/en/lsweek.json
```diff
@@ -4,15 +4,64 @@
...
+			"values": [
+				"en",
+				"de-DE",
+				"de-CH",
+				"..."
+			]
+		},
+		{
+			"name": "timezone",
+			"description": "A datetime object is independent of a specific timezone; it is only an offset in milliseconds from 1970-1-1 00.00:00 UTC (Coordinated Universal Time).\n\nThe timezone only comes into play when you need specific information like hours in a day, minutes in a hour or which day it is as these calculations depend on the timezone.\n\nA timezone must be specified in order to translate the date object to something else. If you do not provide the timezone in the function call, it will default to the timezone specified in the Lucee Administrator (Settings/Regional), or the timezone specified for the current request using the function setTimezone().\n\nYou can find a list of all available timezones in the Lucee administrator (Settings/Regional). Some examples of valid timezones include:\n\n* AGT (for time in Argentina)\n* Europe/Zurich (for time in Zurich/Switzerland)\n* HST (Hawaiian Standard Time in the USA)",
```
shaedrich: The timezone can come into play for all dateparts. Not only hours can be different. For example: 8:00 a.m. here in Germany (= CET) is 11 p.m. the previous day in PST (pacific time zone).

pfreitag: @shaedrich I agree, but it doesn't look like this PR actually changed the description, just the formatting of the json, so I'm going to merge the PR. If you would like to improve the description let's just do that in another PR instead.

pfreitag:
Thanks @CreativeNotice

data/en/lsweek.json
shaedrich: @pfreitag Okay, I'll do that.

shaedrich: See #1199

PR #1200, opened 2019-10-07
Added simple example for fileappend

CraicOverflow89:
Added simple example.


pfreitag:
Thanks @CraicOverflow89 

BTW if you want to do another pull request, you can add `"runnable" : false` to the example, because trycf blocks the file functions. This will remove the run button.

PR #1201, opened 2019-10-07
Created simple exception for getexception

CraicOverflow89:
Added simple exception for the ACF getException method.


pfreitag:
Thanks @CraicOverflow89

PR #1204, opened 2019-10-08
Add example for isLeapYear

bockensm:


pfreitag:
Thanks @bockensm ! If you are looking to do another one, It might be good to have examples that returns true using a hard coded date

PR #1211, opened 2019-10-09
Fix order of "guides" drop-down display

### LINKED ISSUE #1170: fixing order of "guides" drop-down display
As some may have noticed, the "Guides" drop-down is in not in alpha order. I looked into changing it, seeing it is controlled by an array in data/en/guides.json. The issue is that while the values listed there ARE in sorted order, that's the file NAMES holding each guide. The guide TITLES are not reflected in that filename alpha ordering:

`related":["application","cfsqltype-cheatsheet","cgi-scope","closures","coldfusion-versions","datetime","elvis","error-handling","first_class","for","java","member","new","operators","ormsettings","scopes","script","security","security-authentication","security-encryption","security-obfuscation","security-session-management","system-properties","testbox","variables-scope"]`

And of course, one could try to manually figure out what the sort order would be, by the TITLEs, and change that list of array values. 

But I'm wondering if there may be some automated process which **might end up reversing things back to the file-name order**, if/when someone added a new guide to the site.

If that's the case, I just wanted to bring up this issue in the hopes that some consideration might be given to what may be the best way to handle this issue. 

Could this be the solution: **could such an automated indexing operation instead simply sort the values in that JSON "related" array by the TITLE in the files (the first # line in the md file)?** (I tried to find such indexing code but missed it, if it exists.)

shaedrich:
Fix #1170

### DIFF
--- a/Application.cfc
+++ b/Application.cfc
@@ -20,7 +20,8 @@
 					<cfset application.categories[local.cat].items = local.catData.related>
 				</cfif>
 			</cfloop>
-			<cfset application.guides = {}>
+			<cfset _guides = {}>
+			<cfset application.guides = [:]>
 			<cfloop array="#application.index.guides#" index="local.guide">
 				<cfset local.fileObj = fileOpen(ExpandPath("./guides/en/#local.guide#.md"),"read")>
 				<cfset local.title = fileReadLine(local.fileObj)>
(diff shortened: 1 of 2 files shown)


shaedrich:
Tested in trycf, but not yet with CommandBox.

Application.cfc
```diff
@@ -20,7 +20,8 @@
 					<cfset application.categories[local.cat].items = local.catData.related>
 				</cfif>
 			</cfloop>
-			<cfset application.guides = {}>
+			<cfset _guides = {}>
```
pfreitag: Let's use `local.guides` here or `var` scope `_guides`

shaedrich: Sure 👍

shaedrich:
@pfreitag Tested in CommandBox.

shaedrich:
@pfreitag Do you wait for me to test or do you want to test it yourself? In the latter case, is there still any action needed from me?

carehart:
Hi, guys. Thanks for addressing this issue. That said), I'm not yet seeing it reflected on the site. Is it that it will take time before the change propagates to there? To be clear, I did try to ensure I was not seeing cached content. in my browser.

pfreitag:
@carehart - yeah, I have some pretty heavy caching on cfdocs so it might take a day or so to fully flush it.

carehart:
OK, thanks. Totally understandable, and indeed laudable. :-)

PR #1212, opened 2019-10-09
Provided examples and clarification

homestar9:
Provided a few examples of how you can use the `systemOutput()` function to write different object types to the output stream.  I also clarified the description to use the word "object" instead of "string" for clarification.


pfreitag:
Thanks @homestar9

PR #1213, opened 2019-10-11
Added an INSERT example

homestar9:
Who doesn't love a good insert statement?


pfreitag:
I sure love a good insert statement! Thanks @homestar9

PR #1214, opened 2019-10-11
#541 Created simple example for location

CraicOverflow89:
Added simple example - relates to Issue #541.


pfreitag:
Thanks @CraicOverflow89

PR #1219, opened 2019-10-15
Clarify spreadsheet functions that are not supported by Lucee

mjclemente:
Lucee's docs are also being updated to reflect this: https://luceeserver.atlassian.net/browse/LD-126


pfreitag:
Thanks @mjclemente

PR #1220, opened 2019-10-15
Added an example to getSOAPRequest and getSOAPResponse

jordonr:
Provided a working example of using these to functions, and added the `webservice` parameter to `getSOAPRequest`.

### DIFF
--- a/data/en/getsoaprequest.json
+++ b/data/en/getsoaprequest.json
@@ -1,7 +1,7 @@
 {
 	"name":"getSOAPRequest",
 	"type":"function",
-	"syntax":"getSOAPRequest()",
+	"syntax":"getSOAPRequest(webservice)",
 	"returns":"any",
 	"related":["getSOAPRequestHeader","getSOAPResponse","getSOAPResponseHeader"],
 	"description":" Returns an XML object that contains the entire SOAP\n request. Usually called from within a web service CFC.",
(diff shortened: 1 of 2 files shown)


data/en/getsoaprequest.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name":"getSOAPRequest",
 	"type":"function",
-	"syntax":"getSOAPRequest()",
+	"syntax":"getSOAPRequest(webservice)",
```
pfreitag: Good catch here, this is good, however we also need to add it to the `params` array, like this:
```
"params": [
		{"name":"webservice","description":"The description.","required":false,"default":"","type":"string","values":[]}
],

Note: I didn't fill in a `description` above
```

jordonr: Should I also change the type from "string" to "any" as it's listed in the Lucee docs?

pfreitag: yes, should be `any` - good catch

data/en/getsoaprequest.json
```diff
@@ -1,16 +1,27 @@
...
 	"returns":"any",
 	"related":["getSOAPRequestHeader","getSOAPResponse","getSOAPResponseHeader"],
 	"description":" Returns an XML object that contains the entire SOAP\n request. Usually called from within a web service CFC.",
-	"params": [],
+	"params": [
+		{"name":"webservice","description":"A webservice object as returned from the cfobject tag or the CreateObject function.","required":true,"default":"","type":"any","values":[]}
```
pfreitag: I don't think this argument `required` according to adobe docs it is optional

jordonr: I can update, I see that Lucee and the Adobe docs have that.  Also I tried to confirm that by using trycf.com and the following code fails on Lucee and ColdFusion 2018 and ColdFusion 2016:
```
soapURL = "https://graphical.weather.gov/xml/SOAP_server/ndfdXMLserver.php?wsdl";
ws = CreateObject("webservice", soapURL);
zipLatLong = ws.LatLonListZipCode("10001");
//res = getSOAPRequest(ws);
res = getSOAPRequest();
writeDump(res);
```

jordonr:
Hello, just following up if there is anything I missed or need to do still.  Thanks!

pfreitag:
Thanks @jordonr 

BTW if you want to do another PR you can change it to `type="any"` on that argument

PR #1223, opened 2019-10-22
No quotes needed

rscRoman:
No quotes needed for scientific notation


pfreitag:
Thanks @rscRoman

rscRoman:
It is a pleasure to be part of it

PR #1226, opened 2019-10-24
Added Sleep() example

alorne:


pfreitag:
Thanks @alorne

alorne:
Thanks @pfreitag I feel a bit cheap doing all these small ones for hacktoberfest. do you have any slightly larger projects you recommend?

Issue #235, opened 2015-12-11
Server CGI variables


infernocloud:
A page describing Coldfusion's server CGI variables is missing. I was going to create it, but I'm not sure where the best location is and what format it should be. I thought maybe it should go under Guides, but the files in guides/en don't follow the same structure as the files in data/en. This would be tabular data mostly.

The content would of course start with the Adobe reference: https://helpx.adobe.com/coldfusion/cfml-reference/reserved-words-and-variables/cgi-environment-cgi-scope-variables/cgi-server-variables.html

But that page is missing some variables, most relevant to me right now is "LOCAL_ADDR". It would be great to add more of these undocumented CGI variables.

Another thought, instead of a CGI-only page, should this go in a "Scopes" page? This would have CGI, client, server, application, session, and perhaps some others.

pfreitag:
Great Idea -- I think it should go in a guide, I would call it cgi.md then we can later create a guide called scopes.md to link to the various scopes in detail.

The guides are written using Markdown, I will create cgi.md and then you can just edit it with more info.

pfreitag:
Added cgi guide: https://github.com/foundeo/cfdocs/edit/master/guides/en/cgi.md please feel free to complete it!

shaedrich:
This url is wrong. Maybe this could be the one mentioned: https://github.com/foundeo/cfdocs/blob/master/guides/en/cgi-scope.md

PR #1228, opened 2019-10-24
Add some server variables for the CGI scope

jsteinshouer:
I added some of the server variables to the guide per issue #235. If this looks good, I can probably add some client variables as well.


pfreitag:
Great work @jsteinshouer I like that you included examples of possible values

PR #1233, opened 2019-10-25
Added queryClear and did queryRecordCount work

alorne:


alorne:
@pfreitag I'm not sure what the problem is here. I had conflicts in my pull request, so I did a fresh pull from the foundeo/cfdocs repo, fixed the issues then checked them in. The issue shows resolved in github, but seems to be an issue in Travis CI?

alorne:
I was planning on the Tests work to be a separate pull request. But I forgot about the current one and I don't know how to delete the 2 commits, so oh well

TestNamesake.cfc was not working because of the path. Adding in ExpandPath seems to have fixed it.

In that file and the rest of the test files I var scoped the variable in each of them.

shaedrich:
If you  `var` scope variables, to be consistent, you've to access them through `local.varname` :wink:

Maybe this could help you moving the test commits to another branch: https://stackoverflow.com/questions/1628563/move-the-most-recent-commits-to-a-new-branch-with-git

KamasamaK:
You do not have to use the `local` scope prefix to access them, and in the case where they're used in lambda functions you would be referencing the wrong `local` scope anyway and it would break.

shaedrich:
If you don't use local scope in lucee, script execution gets significantly slower because it searches a fair amount of other scopes before it reaches local scope. In lambda functions local variables can be found in `variables` scope.

alorne:
> If you `var` scope variables, to be consistent, you've to access them through `local.varname` 😉
> 
> Maybe this could help you moving the test commits to another branch: https://stackoverflow.com/questions/1628563/move-the-most-recent-commits-to-a-new-branch-with-git

I committed to the branch/pull request without thinking about it. What I couldn't d0 was figure out how to pull the 2 commits back. It doesn't matter though. They can all go together.

alorne:
> If you don't use local scope in lucee, script execution gets significantly slower because it searches a fair amount of other scopes before it reaches local scope. In lambda functions local variables can be found in `variables` scope.

So what is the consensus? or should I abandon the var scoping completly?

shaedrich:
That's up to @pfreitag but your code is valid in the current state. Otherwise, the checks wouldn't have passed.

ddspringle:
> If you don't use local scope in lucee, script execution gets significantly slower because it searches a fair amount of other scopes before it reaches local scope. In lambda functions local variables can be found in `variables` scope.

That's a failure of Lucee, innit? I mean, one would think local would be one of the first scopes to search (esp inside functions, after variables elsewhere) instead of cascading to it, no?

Just an observation - if what you claim is true then a ticket should be raised such that the cascading order is improved in Lucee, if one hasn't already ofc.

KamasamaK:
I'm not familiar with how Lucee places the `local` scope in the `variables` scope, but as far as I know these tests also need to pass in ACF. Regarding it being significantly slower, I would be curious how significant. I found some Lucee documentation on [using scopes explicitly in code](https://docs.lucee.org/guides/developing-with-lucee-server/scope/using-scopes-explicitly-in-code.html) recommending not to explicitly `local` scope in functions.

shaedrich:
That depends on the [`scopecascading`](https://docs.lucee.org/reference/tags/application.html#attribute-scopecascading) setting.

pfreitag:
@ddspringle the `local` scope is searched first within a function as far as cascading goes, but if you don't `var` or `local` scope a variable within a function when you first use it, then it goes in the `variables` scope.

I don't think: 

```
var x = 1;
return local.x;
```

Will be any faster than:

```
var x = 1;
return x;
```
Because if I have `variables.x = 0;` defined, it will always use the `local` scope first, you can write a test case easily to verify: https://trycf.com/gist/58355ffda2d4f13a5e1dd18ec9e67679/lucee5

KamasamaK:
To be clear, I was not saying that the variable would be inaccessible as it is currently, but that you could not explicitly reference its scope from within the lambda. The only _bug_ I'm aware of with referencing outer variables is https://tracker.adobe.com/#/view/CF-4202734.

pfreitag:
Ah, ok makes sense. Sorry - that was my misunderstanding @KamasamaK

pfreitag:
Crikey, now we have a merge conflict because I just merged another pr that added a new function and updated the function index. @alorne let me know if you can fix this

alorne:
@pfreitag I'll give it a shot

alorne:
@pfreitag Fixed! Forgive the mess, I thought I had a problem committing that I clearly did not.

pfreitag:
Thanks @alorne appreciate it!

PR #1244, opened 2019-10-30
Rename E function params - encryptBinary()

shawnoden:
NOTE: I currently only have access to the web interface, so sorry for the multiple PRs.


pfreitag:
Thanks @shawnoden no problem with multiple PRs

PR #1248, opened 2019-10-30
Rename E Function Params - entityDelete()

shawnoden:


pfreitag:
Good additions, thanks @shawnoden

PR #1250, opened 2019-10-30
Rename E Function Params - entityLoad()

shawnoden:
This one has a complex signature with 2 sets of optional params that are mutually exclusive.


data/en/entityload.json
```diff
@@ -1,17 +1,17 @@
 {
 	"name":"entityLoad",
 	"type":"function",
-	"syntax":"entityLoad(entityName [, filterCriteria] [, unique])\n entityLoad(entityName [, filterCriteria] [, sortOrder] [, options])",
+	"syntax":"EntityLoad(entityName [,id | Filter ,unique | Order ,options])",
```
pfreitag: You are changing the case of `entityLoad` to `EntityLoad`, the docs all use `lowerCamelCase` for function names.

shawnoden: I prefer lowerCamelCase too, but NamedParametersColdFusion2018.pdf showed it to have a leading capital. I know we were changing case in the param names, but I think that initial cap may have been an unintentional outcome of the document they generated. If the casing of the function name doesn't matter, and our other function names are primarily lowerCamelCase, then I'll revert that change.

shawnoden: Done.

pfreitag:
Thanks

PR #1253, opened 2019-10-30
Rename E Function Params - EntityNew()

shawnoden:
RE: #898


shawnoden:
NOTE: The `ignoreExtras` param is listed in the CF2018 NamedParameters doc, but I can't find documentation for this param.

pfreitag:
Thanks, we'll fill that in once someone figures out what it does 👍

PR #1255, opened 2019-10-30
Rename E Function Params - EntityToQuery()

shawnoden:
RE: #898


KamasamaK:
@pfreitag I think your JSON validator must have missed an error in this file. The closing quote on line 8 is causing a parsing error -- it should be at the end of line 7.

shawnoden:
I fixed the extra line and created PR #1278, but now the check is failing. I'm not sure how that change could cause it to fail. The JSON validates. Is this a parser issue?

Issue #529, opened 2017-08-31
Multiple signatures


KamasamaK:
See discussion in PR #523 for more. I thought it would be better for this to be an issue instead since this will be a general discussion and not about that specific change.

@pfreitag While that is correct for CFML implementation (ignoring that those functions are really implemented in Java), I would prefer docs that could abstract from that implementation detail and represent what is logically supposed to happen and will be most helpful to a developer. In other words, the caller invoking the function should not be concerned with such implementation details and should only be concerned with how to call it. I support `syntax` optionally being an array for this purpose, but I would want to find a safe way to do this. 

For example, if I say an argument can be either a `string` or `function`, I will have to implement it as `any` out of necessity but if I throw an exception if it's an `array` then is it really `any`? A better example would be conditional arguments. See [`structNew()`](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-s/structnew.html). The `sortOrder` argument can only be used when the second argument is a `string`. Also, even though that argument is optional in this example, I think it would be useful to be able to say that a param is required for a given signature (see `isValid()`).

FYI, in addition to ones mentioned above, here are some other functions I've found that would benefit from multiple signatures: `arraySort`, `createObject`, `listSort`, `replaceListNoCase`, `structSort`, `structToSorted`

shaedrich:
`syntax` change to array is discussed (https://github.com/foundeo/cfdocs/issues/525)[here] as well.

KamasamaK:
An example demonstrating my position would be the **jQuery API documentation**. Even though it is implemented in JavaScript, which does not support function overloading, the documentation does represent multiple signatures. For example, [`.data()`](http://api.jquery.com/data/) has 4 signatures, grouped to have the 2 setters together and the 2 getters together. I am not suggesting that level of grouping, but the distinct signatures as documented is very helpful and represents the abstractions the API provides instead of being beholden to the implementation details of the language.

shaedrich:
To be consistent, we have to decide two things:

First:
If multiple signatures are present AND no argument has influence on another, should every syntax be shown
https://github.com/foundeo/cfdocs/blob/ad46f496f93b97d25e48f713eb0ca4516e95b459/data/en/queryaddrow.json#L4-L10

or could it be pooled into one?
https://github.com/foundeo/cfdocs/blob/ad46f496f93b97d25e48f713eb0ca4516e95b459/data/en/createobject.json#L7

Second:
[...]

PR #1258, opened 2019-11-01
Rename F Function Params - fileGetMimeType()

shawnoden:


data/en/filegetmimetype.json
```diff
@@ -1,12 +1,13 @@
...
 	"type":"function",
-	"syntax":"fileGetMimeType(file, strict)",
+	"syntax":"fileGetMimeType(file | FilePath [, strict])",
 	"returns":"string",
 	"related":[],
 	"description":"Gets the MIME type for the file path/file object you have specified.",
 	"params": [
-		{"name":"file","description":"Name of the file object or full path on disk to the file if strict is set to true. If you do not specify the full path, the file is assumed to be present in the temp directory, as returned by the function getTempDirectory.","required":true,"default":"","type":"any","values":[]},
+		{"name":"file","description":"Name of the file object. If you do not specify the full path, the file is assumed to be present in the temp directory, as returned by the function `getTempDirectory`.","required":true,"default":"","type":"any","values":[]},
+		{"name":"FilePath","description":"Full path on disk to the file.","required":true,"default":"","type":"any","values":[]},
```
KamasamaK: This parameter name should stick with the convention of using lower camel case. Also, I think the `type` should be `string`.

shawnoden: RE: return type for `file` and `filePath` >> Lucee lists it as an `any` type, but I would agree that it should be a `string`. `getTempDirectory()` returns a `string`, and I am not sure if you can pass anything other than a string in that position. 

RE: `FilePath` vs `filePath` >> I also agree that it should stick with the lowerCamelCase convention, but the NamedParametersColdFusion2018.pdf doc shows it as `FilePath`. Is that a type on their part? I'll happily change it if it won't break anything.

KamasamaK: Combined they would be `any` since `file` is `any`, but since you separated them it makes sense that `filePath` would be `string`, not `any`. Side note, separating/multiple signatures has its own issue -- #529.

Regarding case, CFML is a case insensitive language so it won't break anything. This is just a matter of CFDocs convention, so it doesn't matter what case the engine documentation may use.

shawnoden: Gotcha. Can you actually submit a file object into this function rather than just the name of the file? That was what I was going to test. I thought it was just a string name that referenced the file. I'm not sure I've used this function. Years ago, I think I took Pete's "Never Trust a Mime" example to heart. :-)

As far as the multiple signature values, I thought we had previously used a pipe character to delineate when two different param names could be used. But if they are, in fact, two different types rather than just two different names in the same position, then I can see how that could be bothersome. 

RE: the casing, I knew it didn't matter to CF, but I thought one of the reasons for making this changes was to not break implementations that did use CFDocs and ACF2018 param names in a case-sensitive manner. I've been making sure that these changes reflect the casing in the Adobe document. Though I've already found a couple of places where their pdf doesn't match the implementation. Like `hash(...,additionalIterations=1)`. Did we ever get any clarification as to what was the truth source regarding parameters? Per previous notes, I've been using the PDF doc as the truth.

pfreitag: Hmm - do we want to split this into two `params` I think for other function arguments that accept multiple types, we just use type `any` and then specify in the description that it can take a file object or a file path.

KamasamaK: I would say until we can adopt a solution to #529 to keep them combined. Separating without that would cause an issue with my VS Code extension, and I'm not sure about other tools.

shawnoden: So how do you need me to fix it?

shawnoden: @pfreitag @KamasamaK Ping again. Do I need to do anything with this one, or just close it for now?

pfreitag: Hey @shawnoden so I think we either use file or filePath here for the name and then update the description to note that file is for a file object, and filePath is the name for the path named argument.

PR #1273, opened 2019-11-08
Add locale example for lsNumber/Currency/EuroCurrency/TimeFormat

rscRoman:


data/en/lseurocurrencyformat.json
```diff
@@ -1,13 +1,14 @@
 {
 	"name":"lsEuroCurrencyFormat",
 	"type":"function",
-	"syntax":"lsEuroCurrencyFormat(currency, type)",
+	"syntax":"lsEuroCurrencyFormat(currency [, type, locale])",
```
pfreitag: It says that `type` is required so it should not be in square brackets here right?

rscRoman: I'm not shure if type is required.
lsEuroCurrencyFormat(5) works with "currency" only

pfreitag: Yes, according to the adobe docs `local` is the default type, so we should just update the `required` and `default` on the param below. Can you take care of that?

rscRoman: Of course!
I do some updates and commit them.

rscRoman:
@pfreitag 
Is there anything else needed on this Pull request?

KamasamaK:
It looks like the given `values` for `locale` are just suggestions. That field is supposed list all values if enumerable. And I'm guessing `...` is invalid.

rscRoman:
I am pretty sure that `...` is invalid.
Just to be sure that it is accepted for cfdocs, i took the values from [https://cfdocs.org/lsweek](url) as reference.

Issue #1209, opened 2019-10-09
Add new functions/changes in CF2018 Update 5


cfvonner:
### New Array Functions
- ArraySome
- ArrayEvery
- ArraySplice
### New Struct Functions
- StructSome
- StructEvery
### New Query Functions
- QuerySome
- QueryEvery
- QueryAppend
- QueryPrepend
- QueryRowSwap
- QuerySlice
- QueryClear
- QueryRecordCount
- QueryReverse
- QueryInsertAt
### Other Language Updates:
- **GetMetaData** - Changes in Array.GetMetaData struct keys.
- New parameter **_includeEmptyFields_** in fhe following functions:
  - ListAppend
  - List Prepend
- **QueryNew** - Insert rows in a query object as array of .
- **cfchart** - Added attribute **_base64_**.
- **Member functions** - Contains member functions for newly added functions.
- **Array Notation for Strings in ColdFusion** - Access sub-strings using array notation by using positional ranges.
- Regex changes for the following functions:
  - ReFind
  - ReFindNoCase
  - ReReplace
  - ReReplaceNoCase
  - ReMatch
  - ReMatchNoCase
### Arrow functions

See[https://helpx.adobe.com/coldfusion/kb/coldfusion-2018-update-5.html](https://helpx.adobe.com/coldfusion/kb/coldfusion-2018-update-5.html) for more information and links to new/improved functions.

alorne:
I'm going to work on some of the query functions tomorrow. Are there any issues with using comic book characters as my sample data?

srbaxter:
> 
> 
> I'm going to work on some of the query functions tomorrow. Are there any issues with using comic book characters as my sample data?

Sorry @alorne, didn't see this until after I did a couple. I only did queryappend and prepend, so feel free to do the others. I'll move back to the "Other Language Updates" section.

alorne:
No problem @srbaxter

alorne:
Is there any info on QueryClear? I can't seem to get it functioning.

alorne:
> Is there any info on QueryClear? I can't seem to get it functioning.

Same with queryreverse

srbaxter:
@alorne I just went to Adobe's pages and put one of their examples in cfdocs's live pens and it worked for both.

https://helpx.adobe.com/coldfusion/kb/coldfusion-2018-update-5.html

alorne:
[...]

PR #1275, opened 2019-11-13
arrow function

spillsthrills:
#1209


shaedrich:
Hmmm … Arrow is not a real function itself, it's anonymous. So I would rather make a guide instead.

pfreitag:
I agree, lets make it a guide instead.

@spillsthrills to make a guide you just have to create a markdown file in the `guides/en` folder named `arrow.md` and it will become accessible at the URI `/arrow`

pfreitag:
closing this, please resubmit as a guide

PR #1279, opened 2019-11-22
Automatic creation of guide files

rscRoman:
Read and parse data.json to automatically create MarkDown files for following guides:
* ACF only
* Lucee only
* Deprecated
* Discouraged


shaedrich:
I'm not sure if we would consider that guide. The closest thing we have I can think of, are listings.

pfreitag:
@shaedrich I think these are ok in guides for now, since there is no way to add some of the additional metadata notes to listings.

rscRoman:
@pfreitag This is my first Version of the automatic generation. Please give me any feedback if i should rebuild something.

pfreitag:
Looks good to me, what do you think @shaedrich ?

shaedrich:
Besides that, it looks good to me @pfreitag.

pfreitag:
Ok, I'm going to merge it we can improve as needed - looks like handy info @rscRoman thanks for adding it!

rscRoman:
I am happy that I can contribute to this great project.
If there is more that i can do, let me know.

pfreitag:
One suggestion might be to use a `UL` instead of `h3` eg: `* functionName` instead of `### functionName` for each entry, I think it may render a bit cleaner that way.

rscRoman:
Where should i put the "Minimum Version/Deprecated" info?
Looks a little lonely then.

rscRoman:
Made a new Pull request #1280  with an upated view

PR #1284, opened 2019-12-11
Loop from is the beginning, to is the ending

randyclassen:
Swap the from and to descriptions on the cfloop page. The examples and description of index are correct, but the from and to descriptions were reversed.


pfreitag:
Oh wow, good catch there @randyclassen and thanks for fixing it!

PR #1285, opened 2019-12-12
Attribute cacheid not supported by any versions of Lucee or Railo

richardherbert:


KamasamaK:
That's not the only parameter not supported by another engine. When the param description starts with a tag like `CF10+` and no tag for the other engine then it usually indicates that the engine support for it is limited. There are better ways to indicate this, like an `engines` entry under the `params` item, but I don't think the application is set up for that right now.

richardherbert:
I'm happy to change the format if you have an example to follow. I've several others cache functions to update.

richardherbert:
I'll withdraw this PR and replace it with an alternative method of highlighting.

PR #1287, opened 2020-02-02
Get and formfield aren't allowed with lucee

zspitzer:
CFHTTP: All examples crash with Lucee
https://cfdocs.org/cfhttp


pfreitag:
Thanks Zac!

zspitzer:
my pleasure!

PR #1291, opened 2020-02-17
Updating array to include ITEM

ianosullivan:
Updating cfloop for array to include ITEM as well as INDEX


pfreitag:
Thanks guys - sorry for the delay merging

PR #1294, opened 2020-02-25
ArrayEvery added in ColdFusion 2018 u5

aliaspooryorik:
Set minimum_version as `2018,0,05` not sure if that's how you flag the update version


KamasamaK:
To be consistent, we use semver style versions, so it would be `2018.0.5`.

aliaspooryorik:
@KamasamaK thanks for feedback - updated

pfreitag:
Thanks!

PR #1296, opened 2020-02-27
adds offset to list of possible struct keys for param queryoptions

letskillowen:


pfreitag:
Thanks, sorry for the delay in merging!

PR #1297, opened 2020-02-29
Removed unwanted/extra semicolon in script syntax of cfabort

saravanakumar404:
**Removed unwanted/extra semicolon in script syntax of cfabort**

### DIFF
--- a/data/en/cfabort.json
+++ b/data/en/cfabort.json
@@ -2,6 +2,7 @@
 	"name":"cfabort",
 	"type":"tag",
 	"syntax":"<cfabort>",
+	"script":"abort;",
 	"related":["cferror","cfthrow","cfexit"],
 	"description":"Stops the processing of a CFML page at the tag location.\n CFML returns everything that was processed before the\n tag. The tag is often used with conditional logic to stop\n processing a page when a condition occurs.",
 	"params": [


data/en/cfabort.json
```diff
@@ -2,6 +2,7 @@
 	"name":"cfabort",
 	"type":"tag",
 	"syntax":"<cfabort>",
+	"script":"abort;",
```
shaedrich: As far as I can see, your changes do not align with your pr's title: There's an additional semicolon,

saravanakumar404: In cfdocs, cfabort tag script syntax displaying as `abort;;`(having double semicolons) in UI. by using this line I've avoided/ removed extra semicolon. please see the screenshot.
double_semicons

KamasamaK:
This issue causing the extra semicolon is at https://github.com/foundeo/cfdocs/blob/117185026e02672af92fabc725083889eb8ac674/views/doc.cfm#L37-L39

It's appending another `;` even though the `replaceScript` function is already appending a `;`.

saravanakumar404:
> The issue causing the extra semicolon is at
> 
> https://github.com/foundeo/cfdocs/blob/117185026e02672af92fabc725083889eb8ac674/views/doc.cfm#L37-L39
> 
> It's appending another `;` even though the `replaceScript` function is already appending a `;`.

@KamasamaK I can able to remove second semicolon using this line "script":"cfscript;"
please see my screenshot.
image

this is my local result.

pfreitag:
Ok, my thought on this is let's just rely on the `"script"` key and we can remove that bit of logic in the view. So as such I've merged this one.

PR #1298, opened 2020-03-03
Update refindnocase.json

stephen-condon:
Added example for the CF2016+ case using the scope argument

### DIFF
--- a/data/en/refindnocase.json
+++ b/data/en/refindnocase.json
@@ -27,6 +27,12 @@
 			"description": "",
 			"code": "reFindNoCase(\"E\", \"test 123!\")",
 			"result": "2"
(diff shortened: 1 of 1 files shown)


pfreitag:
@stephen-condon thanks for the PR, the build is failing however, might be a syntax error in your json. Let me know if you need help figuring out the problem.

pfreitag:
Hi @stephen-condon thanks for the PR!

If you specify a `result` in the example, our tests will validate that the result you specify is the same as would running your code. So that gets a little tricy when something returns a non complex object, we'll have to improve that at some point so we don't need to serializeJSON it, but thanks for playing along with that quirk!

PR #1299, opened 2020-03-03
Update generatePBKDFKey

### LINKED ISSUE #1293: generatePBKDFKey
According to official CF docs (and use seems to confirm) keySize should be in bits. Cfdocs has "The length in bytes of the key to generate", which should be "The length in bits of the key to generate". It also includes a values section with boolean values underneath.

GeoffCapper:
Change keySize to bits and remove true/false values. Fixes #1293


pfreitag:
Thanks @GeoffCapper

PR #1300, opened 2020-03-04
Updated mime types for Excel

angel-chrystian:
For Excel there are more MIME types for the newest versions of Microsoft Office including the XLSX and open office formats.
Source: https://filext.com/faq/office_mime_types.html


KamasamaK:
There are simply too many MIME types to list them all. Given the purpose of `values` (based on #489), perhaps they shouldn't even be enumerated there.

pfreitag:
Yes, we shouldn't enumerate handy values in `values` it is for a finite list of options, we need another json key, maybe `suggestions` to list handy possible values which IDEs can use to provide autocomplete suggestions.

KamasamaK:
@pfreitag Did you change your mind on this?

pfreitag:
@KamasamaK nope, I merged it, and edited to add `suggestions` key.

PR #1305, opened 2020-03-23
Update cfexecute.json

DomTheBass:
Example shows success attribute which errors e.g. in Lucee "Attribute success is not allowed for statement execute"


pfreitag:
Thanks!

PR #1309, opened 2020-05-13
Update application-cfc.json

eyk:
The type of this.secureJSONPrefix is string (not boolean), therefore the values:  true, false make no sense here.


pfreitag:
Thanks @eyk

PR #1313, opened 2020-05-26
Cfcookie fix examples to use encodevalue

GeoffCapper:
Cfcookie examples used "encodedvalue" rather than "encodevalue".


pfreitag:
Thanks @GeoffCapper

PR #1314, opened 2020-05-29
lsParseNumber add "locale" param

ecellini:
Hello,
I have add "locale" param on lsParseNumber  function


pfreitag:
Thanks @ecellini

PR #1316, opened 2020-06-15
Add samesite param with abbreviated Adobe description

peterrother:
More information available from Adobe:

- https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-c/cfcookie.html
- https://helpx.adobe.com/coldfusion/kb/coldfusion-2018-update-9.html


pfreitag:
Thanks @peterrother

PR #1317, opened 2020-06-20
removed e at the end of hours

randy-johnson:
``` 
Time zone in hourse of offset in ISO 8601 format
```
to
``` 
Time zone in hours of offset in ISO 8601 format 
```


pfreitag:
Thanks @randy-johnson

PR #1320, opened 2020-07-21
Update cfthread.json Add attribute

Vintic:
Add not specified attribute, that pass values to thread in ARGUMENTS scope.


pfreitag:
I think this would be better to document in the description, rather than in the tag attributes, since these definitions may be used for more than just display purposes.

pfreitag:
closing this pr, feel free to submit again in the description rather than by creating a new argument

PR #1322, opened 2020-08-01
Added the examples about directorycopy,empty,lsisnumeric changes.

Dharcf:
Hi Foundeo,

I've added the examples about directoryCopy, Empty, Lsisnumeric. Please check it and confirm. Thanks.

### DIFF
--- a/data/en/directorycopy.json
+++ b/data/en/directorycopy.json
@@ -23,7 +23,7 @@
 		{
 			"title":"Simple DirectoryCopy Example",
 			"description":"Copy directory from one place to another.",
-			"code":"<cfoutput>#directoryCopy(expandPath(\"./mySourceDirectory\"),expandPath(\"../MyDestinationDirectory\"))#</cfoutput>",
+			"code":"directoryCopy(expandPath(\"./mySourceDirectory\"),expandPath(\"../MyDestinationDirectory\"))",
 			"result":"",
 			"runnable":false
 		}
--- a/data/en/directorycopy.json
+++ b/data/en/directorycopy.json
@@ -18,6 +18,15 @@
 	},
 	"links": [
 
+	],
+	"examples": [
+		{
+			"title":"Simple DirectoryCopy Example",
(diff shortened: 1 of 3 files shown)


data/en/directorycopy.json
```diff
@@ -18,6 +18,15 @@
 	},
 	"links": [
 
+	],
+	"examples": [
+		{
+			"title":"Simple DirectoryCopy Example",
+			"description":"Copy directory from one place to another.",
+			"code":"<cfoutput>#directoryCopy(expandPath(\"./mySourceDirectory\"),expandPath(\"../MyDestinationDirectory\"))#</cfoutput>",
```
pfreitag, line 26: It is not necessary to add the `<cfoutput>` and `#` symbols around a single function call, one liner. It will be wrapped automatically, so to clarify you can just use this instead:

"code": "directoryCopy(expandPath(\"./mySourceDirectory\"),expandPath(\"../MyDestinationDirectory\"));"

data/en/lsisnumeric.json
```diff
@@ -17,6 +17,20 @@
 	},
 	"links": [
 
+	],
+	"examples": [
+		{
+			"title":"Numeric lsisnumeric",
+			"description":"",
+			"code":"<cfoutput>\r\n\t#lsIsNumeric('467878')#\r\n</cfoutput>",
```
pfreitag, line 25: Same issue as above, let's just use `lsIsNumeric('467878');` as the code

data/en/lsisnumeric.json
```diff
@@ -17,6 +17,20 @@
...
+			"result":"YES"
+		},
+		{
+			"title":"String lsisnumeric",
+			"description":"",
+			"code":"<cfoutput>\r\n\t#lsIsNumeric('Ten10')#\r\n</cfoutput>",
```
pfreitag, line 31: cfoutput, and pound can be removed here as well. Otherwise everything looks good.

PR #1323, opened 2020-08-01
Added the example about gettickcount, gettimezone,ispdffile,lsdayofwe…

Dharcf:
Hi Foundeo,
Here I've added the examples about gettickcount and some other functions. Please check it and confirm that. Thanks.


pfreitag:
@Dharcf the case of function names is inconsistent, eg `GettickCount` and `getTickCount` can you update these to use consistent case, eg `getTickCount`

PR #1338, opened 2020-08-13
Added links for applicationStop function.

saravanakumar404:
Added useful links for applicationStop function.


pfreitag:
@saravanakumar404 The link to Ben Nadel's blog is good, but we already have the link to the Adobe Docs under engines. Can you remove that one?

saravanakumar404:
> @saravanakumar404 The link to Ben Nadel's blog is good, but we already have the link to the Adobe Docs under engines. Can you remove that one?

@pfreitag As per your comments, I've removed an Adobe Docs link and make a commit, check it.

PR #1346, opened 2020-09-02
Fix bad example for Lucee's nullValue() function

atuttle:
Current example doesn't use the method it's supposed to be demonstrating.


pfreitag:
Thanks @atuttle

PR #1348, opened 2020-09-24
Fix terminateOnTimeout docs for cfexecute

atuttle:
I'm not sure about the listed version numbers, but it looks like there's some confusion.

The ACF docs history section only references a change made in ACF6.1 (adding the `variable` attribute), with no mention of initial addition of the method.

Likewise, the Lucee docs don't mention when the method was added.

For that reason, I've removed the version numbers referenced here. Hopefully the feature is old enough that it doesn't matter any more. :)


pfreitag:
Hi @atuttle 

It looks to me like `terminateOnTimeout` is only supported in Lucee, so I think we want to keep `Lucee4.5+` in the description because that will trigger a label to show up. 

The compatibility note should probably be on ACF, and say `terminateOnTimeout` is not supported on ACF right?

KamasamaK:
Regarding when this was added to ColdFusion, in `45langref.pdf` pg. 29 (see [CFML Repo](https://drive.google.com/drive/folders/0Bwmlt4TArPS5dlVSQVAtV2R5M2M) en docs) it is indicated that `cfexecute` is a new tag.

atuttle:
Made the requested changes

PR #1351, opened 2020-10-02
#898 - Update getbasetagdata.json

shawnoden:
#898


shawnoden:
The parameter names defined inside the function differ from current documentation. ACF = "ancestorLevels"; Lucee = "instancenumber ".  Verified "instance" is correct in the function itself.

PR #1353, opened 2020-10-02
Fix name of queryExecute options attribute per #1345

### LINKED ISSUE #1345: incorrect param name for queryExecute
queryExecute lists the third param as 'queryOptions', but it is actually 'options'. Not worth the trouble for me to get set up to edit, but if anyone would like to update it while in there, it would be great.

Per CF error in CF2018:
Detail | The Valid Parameter for function QUERYEXECUTE are [sql, params, options]

atuttle:
Resolves #1345

### DIFF
--- a/data/en/queryexecute.json
+++ b/data/en/queryexecute.json
@@ -1,7 +1,7 @@
 {
 	"name":"queryExecute",
 	"type":"function",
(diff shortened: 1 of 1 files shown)


pfreitag:
Thanks @atuttle

PR #1354, opened 2020-10-03
Update listsome.json

CraicOverflow89:
Hopefully we are using array nowadays instead of string lists but here's an example for `listSome`.


pfreitag:
Thanks!

PR #1355, opened 2020-10-03
Update applicationstop.json

CraicOverflow89:
Difficult to find a modern use case for this gem but here's a basic example of the `applicationStop` function.


pfreitag:
Thanks @CraicOverflow89

PR #1359, opened 2020-10-16
not just null

savaticus:
Updated to warn other what I just discovered, and that is that if the value being tested is false, then the value on the right is used. This resulted in a true / false variable with a default of true always being true.

### DIFF
--- a/guides/en/elvis.md
+++ b/guides/en/elvis.md
@@ -33,6 +33,8 @@ Similarly, you can use this operator for Struct:
 
     securityNumber = securityStruct['Joe'] ?: "";
 
+NB: If the value tested === false, then the expression on the right will be evaluated.
+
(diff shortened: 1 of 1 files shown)


KamasamaK:
Yes, unfortunately https://tracker.adobe.com/#/view/CF-4198933

I'm not sure how Lucee handles it.

pfreitag:
@savaticus maybe it would be good to add a link to that bug that @KamasamaK pointed out in the guide, do you think you could add that as well? Also would be worth checking lucee, can you run a quick test on trycf.com to see if it is the same or not?

savaticus:
@pfreitag I have done both.

pfreitag:
Thanks @savaticus and @KamasamaK for your input!

PR #1360, opened 2020-10-17
issue 1173: note added to cfschedule parameter

brianconner:
note added to cfschedule parameters as described in issue 1173


pfreitag:
Thanks @brianconner

PR #1361, opened 2020-10-25
Fix for: [MAJOR] Markdown can't parse tables? #652

jmharrin:
Replaced txtmark with flexmark JAR and updated code.  MD conversion now supports tables "pipe" syntax.


jmharrin:
FYI...When merging this pull request requires server restart or application reinitialization to clear out the old .jar file reference from memory.

pfreitag:
Not opposed to switching to flexmark - as long as all the testing is done on the existing markdown pages. So if you want to re-open a PR with just that change feel free, there are some datetime changes in this PR that I don't think are safe to merge. Closing this one.

PR #1364, opened 2020-11-19
Fixed wrong fileUpload argument name "onConflict", correct name is "nameConflict"

netcom-severs:


KamasamaK:
It shows `onConflict` in http://cfdownload.adobe.com/pub/adobe/coldfusion/2018/publicBeta/NamedParametersColdFusion2018.pdf.

Have you verified the proper named parameter in CF2018+?

netcom-severs:
In the Coldfusion docs the parameter is "nameConflict" [https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-e-g/fileupload.html](url) 
Same in the lucee docs: [https://docs.lucee.org/reference/functions/fileupload.html](url)

I have testet the with commandbox 2018, indeed it throws an error:  

> Parameter validation error for the FILEUPLOAD function. 
> The Valid Parameter for function FILEUPLOAD are [destination, filefield, mimetype]

`<cfif 0>
    <cfset fileUpload(destination="tmp", nameConflict="error" )/>    
</cfif>
`

I will update the PR.

netcom-severs:
I can't decide which is the most obvious way to show the incompatibility in the parameter name between lucee and coldfusion.

Duplicating the the hole parameter marking one as lucee only and the other as coldfusion only?
A hint in the parameter description?
Or a note in the engines section?

Any suggestions?

KamasamaK:
Because Lucee supports aliases, the way to solve the incompatibility is to have them add an alias. They are very willing to fix compatibility issues, and an alias is a simple fix. Because ACF is less flexible, it's best to use their name for the parameter in CFDocs.

netcom-severs:
https://luceeserver.atlassian.net/browse/LDEV-3156

pfreitag:
Thanks @KamasamaK for taking care of this one, and thanks @netcom-severs for pointing this out and raising the issue on Lucee.

PR #1368, opened 2020-12-14
Several changes to toString() function

shaedrich:
– More descriptive title for example
– Made one example runnable and removed that property from another one
– Added warning to the latter
Thanks to @MordantWastrel via https://cfml.slack.com/archives/C06T99N9G/p1606343827458100?thread_ts=1606343827.458100&cid=C06T99N9G

### DIFF
--- a/data/en/tostring.json
+++ b/data/en/tostring.json
@@ -19,10 +19,11 @@
 	"links": [],
 	"examples": [
 		{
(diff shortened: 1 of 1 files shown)


pfreitag:
Thanks @shaedrich

shaedrich:
You're welcome @pfreitag

PR #1375, opened 2021-01-25
Added ability to set default TryCF engine per example

euans:
Submitting this as a separate PR because I don't know if this is a feature you'll find useful.  This feature adds the ability to set the default engine on an example so the user doesn't have to select it if it's example is engine specific (ie: CF2021 structNew case-sensitive feature).


pfreitag:
Going to close this one for now, it is a good idea but I think it can be done without passing the engine in the url, we can just read the example json in trycf.cfm for example. If you want to resolve the merge conflicts and fix that feel free to resubmit, I'm just closing this to keep things organized.

PR #1376, opened 2021-02-11
updated to point to more than just encodeforhtml

carehart:
and also removed parens after some named functions, so that the feature here to auto-link to them would work


pfreitag:
Thanks @carehart

PR #1378, opened 2021-03-18
ColumnType "timestamp" was missing

JamoCA:
ColumnType `timestamp` is supported in every cfml application server.  Adobe incorrectly states in their [documentation](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-m-r/queryaddcolumn.html) that "Date (can include time information)", but it doesn't and hasn't since CF11. (Specifying `date` ignores all time values.)


pfreitag:
Good catch @JamoCA thanks!

PR #1379, opened 2021-03-23
Fix references to struct with query

cfvonner:
Initial syntax and `query` callback parameters refer to "struct" instead of "query".


pfreitag:
Thanks @cfvonner

PR #1380, opened 2021-03-25
new JVM flag affecting Application.cfc behavior

softwareCobbler:
CF2018 HF11 and CF2021 HF1 fixed a bug with cfinclude resolution from inside application.cfc; the fix requires a JVM flag, which ought to be documented somewhere

https://helpx.adobe.com/coldfusion/kb/coldfusion-2018-update-11.html


pfreitag:
@softwareCobbler Thanks for this - I agree it should be documented, I think it would make more sense if it were in the `links` key of the json rather than the `examples` section. The `code` would go in the `description` key of the link. Would you be able to make that change?

PR #1381, opened 2021-04-01
Typo: Sort should use "desc" instead of "dec".

JamoCA:
Typo: Sort should use "desc" instead of "dec".


pfreitag:
Thanks @JamoCA

PR #1396, opened 2021-06-17
Fix member syntax example

GaryStanton:
Existing example shows `addRow` instead of `addColumn` and misses an array argument.


pfreitag:
Thx @GaryStanton

PR #1398, opened 2021-07-05
Update queryclear.json (missing params?)

softwareCobbler:
I don't know if this is a bug in this project (is the JSON schema intended to always have a `params`?), or in KamasamaK's plugin; but the absence of a `params` field is unexpected in the KamasamaK plugin, and is met with an exception on a failing read in `CfDocsDefinitionInfo::toGlobalFunction`

so this could be fixed here, but might not be correct if it doesn't matter to CfDocs that there is no `params` field.


pfreitag:
Thanks, I think in general we should expect to have a params array, even if empty.

PR #1402, opened 2021-07-29
Update listcontains.json

tomchiverton:
Because this looks for substrings, unlike listFind, a better example would use that.


pfreitag:
@tomchiverton Agreed, thanks!

PR #1404, opened 2021-09-22
Fixed example CFML; Add note re:compatibility

JamoCA:
Multiple syntax errors. Code appears to work with Adobe ColdFusion CF11 - CF2021, but not Lucee. (Tested using TryCF.com.)


pfreitag:
Thanks @JamoCA

Issue #1366, opened 2020-12-01
CF2021 New Functions / Features / Improvements 


shawnoden:
CF2021 has a bunch of changes and improvements that need to be documented here. This could easily become its own project, since there are far too many changes for a single Issue. I'm throwing this up here as kind of a bookmark.

Adobe lists the New And Changed Functions at https://helpx.adobe.com/coldfusion/cfml-reference/new-changed-functions-coldfusion-2021.html

Several of the Improvements, like Identity Operators, Dynamic Switch/Case or Rest/Spread, could probably benefit from their own Guides. 

# For the new / changed functions: 

## Tags added to CFScript:
* ~SaveContent~ #1408 

## String Functions:
* ~StringSort~ #1410
* ~StringSome~ #1410
* ~StringReduce~ #1410
* ~StringMap~ #1410
* ~StringFilter~ #1410
* ~StringEvery~ #1410
* ~StringEach~ #1410

## Encryption Functions:
* ~GenerateBCryptHash~ #1411
* ~GenerateSCryptHash~ #1411
* ~VerifyBCryptHash~ #1411
* ~VerifySCryptHash~ #1411
* ~Encrypt/Decrypt/EncryptBinary/DecryptBinary - Added support for authentication encryption~ #1413 

## Encoding:
* ~BinaryEncode/BinaryDecode - Added Base64Url encoding~ #1413 

## Structs:
* ~StructIsCaseSensitive~ #1410
* ~StructNew - added two new struct types (ordered-casesensitive, casesensitive)~ #1427

## Arrays:
* ~ArrayPush~ #1408 
* ~ArrayPop~ #1408 
* ~ArrayShift~ #1408 
* ~ArrayUnShift~ #1408 

## New ReduceRight methods:
* ~ArrayReduceRight~ #1412
* ~StringReduceRight~ #1412
* ~ListReduceRight~ #1412

## Time zone functions
* ~GetTimeZone~ #1408     
* ~SetTimeZone~ #1408 
* ~ClearTimeZone~ #1408 

## New flags and features in Application.cfc 
* ~this.timeZone~ #1581
* ~MongoDB settings~ #1581
* AWS features
* ~SAML features~ #1581
* Azure features
(also verify enhancements made in CF2018)
(https://helpx.adobe.com/coldfusion/cfml-reference/application-cfc-reference/application-variables.html)

## Replace/ReplaceNoCase 
* added CallBack support
* new parameter - "start"
[...]

PR #1410, opened 2021-10-08
Cf2021 adds part2 - String Functions - #1366

shawnoden:
Added new CF2021 String functions and Struct function. #1366


shawnoden:
Also, regarding the `functions.json` and other files, there is a function in Utilities, I believe, that will add new functions based on existence of a .json file. Does this still work and do you run it when you push to Master, or do I need to be adding new functions to their required lists? @pfreitag

shawnoden:
Sorry. Just meant to comment. Didn't mean to comment and close.

PR #1415, opened 2021-10-15
Update arrayeach.json

Daemach:
The "each" member functions return the original array on Lucee so they are chainable.  In all of the years of using cfdocs as my primary documentation source, I didn't find this out until last week...


pfreitag:
Feel free to open a new PR with an engine note.

PR #1416, opened 2021-10-17
Ignore Named CommandBox Server Configs

shawnoden:
Added wildcard flag to .gitignore to ignore any server-*.json file created by CommandBox


shawnoden:
@pfreitag Also, none of my previous PRs are being picked up for #Hacktoberfest. Is there something I need to do or that needs to change?

PR #1417, opened 2021-10-18
boolean types were missing true,false accepted values

webmandman:


webmandman:
In the last commit I chagned the regex pattern for allowed page names. Names now allow the period, underscore and forward-slash, plus I removed the hard coded pattern for Application.cfc because once I added the period "Application.cfc" would pass as well.

webmandman:
all the previous regex pattern allowed the name to start with lowercase or uppercase. I know the name description says to start the tag or function name with a lowercase, unless its "Application.cfc". Necessary to start with a lowercase?

webmandman:
the pattern I committed should be fine since I found other names that start with an uppercase, like the "UI Tags" page. I think the only way to make sure tags and functions start with a lowercase is to create a seperate schema for only tags and functions, and another schema for other types of pages, basically any other page that is not a tag or function page.

PR #1418, opened 2021-10-19
Fixing the JSON schema validation issues on CFdocs

mikegrogan:
Everything validates now except index.json. That one seemed pretty unique and I didn't want to mess with it. I did the best I could on the rest of the types, etc. A few of the changes might need to be reviewed, but I think most of it is solid.

A few types could have had multiple results. I just chose one of the valid responses if I had too. A few of the types should have been something like an object, or instance of an object, etc. and there wasn't a valid type of that, so I did the best I could.

I did alter the name pattern slightly to include "Coldfusion 4.5..." and "cf_socialplugin".

### DIFF
--- a/data/cfdocs.schema.json
+++ b/data/cfdocs.schema.json
@@ -170,7 +170,7 @@
       "title": "Name",
       "description": "The name of the tag or function. Start with lowercase, except Application.cfc.",
       "type": "string",
-      "pattern": "^([a-zA-Z0-9 \\. _ -]+|Application\\.cfc)$"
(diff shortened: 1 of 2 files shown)


mikegrogan:
Everything validates still except index.json.

pfreitag:
Awesome thanks - I merged this one, but I did notice one minor thing I'd like you to change. This way you can get some more PR's in which count towards hacktoberfest if you are interested...

Some of the type="timestamp" were changed to type="function" - these should be type="numeric" because the result of `getTimeSpan()` function call is just a number (of seconds).

PR #1419, opened 2021-10-19
Provided example for CFXML

homestar9:


pfreitag:
Yeah @homestar9 - you can ignore the tests for now - working on fixing those at the moment. Never gonna let you down! 

Thanks for contributing!

PR #1421, opened 2021-10-19
Createtimespan returns numeric. updated types to reflect this.

mikegrogan:


pfreitag:
Thanks - I also updated the CI script to exclude index.json since it has a different schema - so you can see we have a green checkmark there now!

mikegrogan:
Awesome!

PR #1440, opened 2021-11-23
add script example

carehart:
Also tweak a typo even in the original/tag example.

### DIFF
--- a/data/en/cfdocument.json
+++ b/data/en/cfdocument.json
@@ -58,10 +58,16 @@
 		{
 			"title":"Dynamic PDF Served to Browser",
 			"description":"PDF is sent back to the browser instead of saving to a file. The cfheader tag is used to suggest a file name to the browser. The cfcontent tag is used to set the Content-Type response header (that could also be done using cfheader) which tells the browser that it is a PDF file.",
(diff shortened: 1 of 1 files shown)


pfreitag:
Thanks @carehart looks great, only issue I see is that you have some `<br>` tags in the example description, can you remove those?

carehart:
Well, they are in the descriprion, and it "worked" to create a paragraph break. Do you prefer another solution for that?

pfreitag:
Yes, if you use `\n` instead of `<br>` it will replace it with a br when it renders as HTML.

There are some other projects using this cfdocs data (such as the vs code editor) so I prefer not to have HTML in the doc data when possible.

carehart:
Btw, by "work" i mean i ran the app via commandbox and the html rendered in the description, without any negative effect on the json parsing. Is it that you maybe feared that, if I'd not tested it? 

Or do you just prefer not to have any html at all in the example descriptions?

pfreitag:
Yep, I realize that it _works_, but I just prefer not to have any HTML at all in the json. There are _checks_ that run automatically to make sure that the JSON is valid, and it would have shown up here where it says "All checks have passed", so I wasn't worried about that.

I will go ahead and merge this, and then change it to the `\n`

carehart:
Awesome. Thanks, Pete, and for the clarification. Will try to remember that going forward,.of course.

PR #1443, opened 2021-12-14
Improve writeBody function and normalize case of other functions

michaelalandawson:
I added more information to the writeBody function.  I found that the return value is boolean in that the function returns "YES" if successful.  I don't know of a way to have it return "NO", however.  If the return is not "YES", then the only other result is an error.  So, technically, it returns a boolean, but in this instance, the return value is practically useless.

I also normalized the use of "Metadata" and "EXIF" within function names.


pfreitag:
good work @michaelalandawson Thanks

PR #1453, opened 2022-02-14
cfimport example

csperando:
Addresses issue #541 

* Included a basic example of using `cfimport` to include a custom tag.
* Added `cfmodule` as a "related" tag.


pfreitag:
Thanks @csperando

PR #1461, opened 2022-02-23
cfqueryparam: Provided testing and formatting for cfscript example #2

Wolfgang-Falk:
Per my previous addition to this page, I have now updated that example to better reflect existing examples for cfqueryparam and its equivalent in cfscript usage. Now, the query executes with a given exampleData query provided, and the writeOutput() statement is also included to have the expected result occur as intended. I also formatted the example like the others on that page for improve legibility.


pfreitag:
@Wolfgang-Falk - I think you are missing the `{ dbtype="query" }` in this example.

data/en/cffunction.json
```diff
@@ -10,7 +10,7 @@
...
-		{"name":"output","description":"Specifies under which conditions the function can generate HTML output.\n","required":false,"default":"","type":"boolean","values":[true,false]},
+		{"name":"output","description":"Specifies under which conditions the function can generate HTML output.\nLucee These values can be defined by "Yes" for true or "No" for false.\n","required":false,"default":"","type":"boolean","values":[true,false]},
```
KamasamaK, line 13: This should be part of a separate PR.

But also, this change is not right. First, the syntax is wrong and breaking the JSON. The `"`s need to be escaped in a JSON string. Second, using "Yes" and "No" for booleans should work in all CFML engines, not just Lucee. The change that should be made here is removing `true` and `false` from `values` since that should only be used for enumerated values.

Wolfgang-Falk: I apologize. I was testing with the documentation on the cffunction page today. and trying the out CF and Lucee notations. At the time, I noticed that Lucee documentation made a mention of only `Yes` and `No` as options, so I was trying it out. I failed to realize that would effect the pull request at the time. I also received the error in JSON, and I plan to fix it. For now, I have removed that addition and will look more into branches and local testing.

Wolfgang-Falk:
Thank you, I have made the changes to the request. I would like to reiterate the change to cffunction was due to testing and saving the test to the master branch and version with the pull request. I have since reverted those change to that page's JSON file.

PR #1462, opened 2022-02-25
cfretry

csperando:
Issue #541 
* added a basic tag and a basic script example for cfretry
* added cfthrow as a related function


pfreitag:
Thanks @csperando - on the cfscript example can you take out the `<cfscript>` tags, we will detect that it is script and allow it to run accordingly. Also in script mode instead of using `LE` operator you can use `<=`

csperando:
@pfreitag changes made

pfreitag:
Thanks @csperando

PR #1464, opened 2022-03-03
Added example of for...in iteration on a list

bockensm:
The comment about not being able to use this syntax on a list is incorrect for any reasonably modern CFML dialect, so I replaced it with an example.

### DIFF
--- a/guides/en/for.md
+++ b/guides/en/for.md
@@ -33,9 +33,16 @@ The above outputs `AB`
 	
 The above example would output `FordDodge`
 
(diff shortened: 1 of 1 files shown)


pfreitag:
Good point - do we know what version this was added in? it might be nice to add a note like: `CF10+` or something.

pfreitag:
works as far back as CF10, not sure if it was added in CF10, perhaps. It doesn't work on Railo 4.2, but works on Lucee 4.5

pfreitag:
Thanks @bockensm if you figure out what version CF added it, feel free to add it.

PR #1466, opened 2022-03-06
w/W flags for week in month/year were reversed

JamoCA:


pfreitag:
Nice catch, thanks @JamoCA

PR #1467, opened 2022-04-07
Added "expandPath"

JamoCA:
The function detail page indicated it was a filesystem function, but it wasn't listed with the other functions on the filesystem functions page.


pfreitag:
Thanks @JamoCA

PR #1472, opened 2022-04-29
add copy to clipboard feature

scottsteinbeck:


scottsteinbeck:
Screen Shot 2022-04-29 at 10 05 43 PM

views/doc.cfm
```diff
@@ -288,6 +288,9 @@
 									<button class="example-btn btn btn-default" data-name="#encodeForHTMLAttribute(LCase(data.name))#" data-index="#example_index#"><span class="glyphicon glyphicon-play-circle"></span>&nbsp; Run Code</button>
 								</div>
 							</cfif>
+								<div class="pull-right" style="margin-right:10px">
+									<button class="copy-btn btn btn-default" onclick="copyTextToClipboard('#encodeForHTML(ex.code)#')" ><span class="glyphicon glyphicon-duplicate"></span>&nbsp; Copy Code</button>
```
KamasamaK, line 292: I feel like "glyphicon-copy" would be better here.

scottsteinbeck: updated

KamasamaK:
I am seeing some JavaScript errors on this because the string being passed to `copyTextToClipboard` is not being encoded for JavaScript, so a single quote or new line will break it.

KamasamaK:
I am getting JavaScript errors whenever I try this because the string being passed to `copyTextToClipboard` is not being encoded for JavaScript, so a single quote or new line will break it.

@scottsteinbeck @pfreitag Can you confirm?

KamasamaK:
Never mind. I think #1488 probably fixes this.

PR #1473, opened 2022-04-30
add open graph meta to site

scottsteinbeck:


utilities/openimage.cfm
```diff
@@ -0,0 +1,131 @@
...
+		name:"CF Docs",
+		type:"All",
+		description:"UltraFast CFML Documentation Reference.",
+		syntax:"",
+		member:"",
+		script:"",
+		engines:{}
+		
+	}
+	if(url.keyExists('name') && FileExists(ExpandPath("../data/en/#url.name#.json"))){
```
pfreitag, line 13: Need to add something above here to sanitize the url.name variable to include it is only a simple string such as: `url.name = ReReplace(url.name, "[^a-zA-Z0-9_-]", "", "ALL");`

scottsteinbeck: Will do

PR #1474, opened 2022-04-30
Update cfmodule.json

KrunchMuffin:
added cfscript example and removed that script is Lucee only

### DIFF
--- a/data/en/cfmodule.json
+++ b/data/en/cfmodule.json
@@ -2,6 +2,7 @@
 	"name":"cfmodule",
 	"type":"tag",
 	"syntax":"<cfmodule>",
+	"script":"cfmodule(template=\"nameoftemplate.cfm\",attributeCollection{});",
 	"script": "module name=\"cached\" {};",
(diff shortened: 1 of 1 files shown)


data/en/cfmodule.json
```diff
@@ -2,6 +2,7 @@
 	"name":"cfmodule",
 	"type":"tag",
 	"syntax":"<cfmodule>",
+	"script":"cfmodule(template=\"nameoftemplate.cfm\",attributeCollection{});",
```
KamasamaK, line 5: Isn't `attributeCollection` supposed to be followed by `=`?

KrunchMuffin:
Yup, that's what I get for copying from adobe website.

data/en/cfmodule.json
KrunchMuffin: Yup, that's what I get for copying from adobe website.

pfreitag:
Thanks @KrunchMuffin

PR #1475, opened 2022-04-30
Add recent items menu to top menu in cfdocs

scottsteinbeck:
add javascript to store recent items in local storage and use it to create recent menu items


scottsteinbeck:
Screen Shot 2022-04-29 at 10 05 16 PM

shaedrich:
@scottsteinbeck  When I opened your PR, I first thought "recent" referred to "recently created" or "recently updated". So maybe, the title should be "recently viewed" to avoid that kind of confusion.

scottsteinbeck:
Yeah I like that idea. What are your thoughts on actually displaying them not as a top menu drop-down. But instead as a secondary nab list it is right under the top nav so the items show horizontally

scottsteinbeck:
Updating the UI for this one, I think the submenu option is going to be a lot more accessable.
Screen Shot 2022-05-03 at 5 00 44 PM

shaedrich:
Interesting. On a gut level, I'd put it into a sidebar on the right or the bottom but in general, I think, that's a good idea. But after all, it comed down to @pfreitag's opinion on that

scottsteinbeck:
Screen Shot 2022-05-04 at 9 47 56 AM

PR #1476, opened 2022-04-30
add example dialog, add script to genereate JSON output for PR

scottsteinbeck:


scottsteinbeck:
Screen Shot 2022-04-30 at 1 59 49 PM
Screen Shot 2022-04-30 at 1 59 17 PM

views/doc.cfm
```diff
@@ -312,6 +364,8 @@
 		</cfif>
 	</div>
 </cfoutput>
+<script src="#request.assetBaseURL#trycf.js"></script>
```
pfreitag, line 367: I don't think it is necessary to load this JS file here

pfreitag:
Thanks @scottsteinbeck

pfreitag:
This works great @scottsteinbeck - one tweak I just thought of though is maybe the link to the github json file should be `target="_blank"` because if you have entered stuff in the form, and haven't copied the json yet you might loose your work.

scottsteinbeck:
yeah thats a great idea. ill update the branch

pfreitag:
One other small tweak, we should only show this one function and tag pages, there is a listing type such as: https://cfdocs.org/cf2021 which is showing the Add Example, but we shouldn't add it there.

The json has a `"type":"listing"` on that page, but we should probably only show Add Example if `type == "tag" || type == "function"`

shaedrich:
Awesome feature! 👏  Would it be helpful to validate if the expected result matches actual one and the code doesn't throw an exception?

pfreitag:
@shaedrich yeah, it would be handy, but I don't want to allow code execution on the server. The github actions should test this once they submit the PR.

scottsteinbeck:
I did have an idea to allow them to import the code from TryCF Since it gets saved in a gist

shaedrich:
@pfreitag I see your point. Was just thinking, that unnecessary fixup commits could be avoided that way.

PR #1477, opened 2022-05-02
add flexmark libs and replace markdown convert function

scottsteinbeck:
I have reviewed all guides to confirm proper markdown to HTML conversion with the flexmark java libraries


pfreitag:
oh ... disregard that last comment I misread the diff

PR #1478, opened 2022-05-03
enable prettify to highlight syntax on guides

scottsteinbeck:
This was missing due to guides not having examples so the code didn't know to load the libraries, plus since the code examples were in markdown, the code blocks were missing the `.prettyprint` class. 

This was easily fixed just by applying the class before running the pretty print function and adding `request.hasExamples = true` for the guides pages
code highlight sbs

### DIFF
--- a/doc.cfm
+++ b/doc.cfm
@@ -3,6 +3,8 @@
 <cfif url.name IS "index">
 	<cfset data = {name="CFDocs", description="Ultra Fast CFML Documentation", type="index"}>
 <cfelseif FileExists(ExpandPath("./guides/en/#url.name#.md")) OR url.name is "how-to-contribute">
+
+	<cfset request.hasExamples = true>
 	<cftry>
 		<!--- convert md to HTML --->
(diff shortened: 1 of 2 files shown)


doc.cfm
```diff
@@ -3,6 +3,8 @@
 <cfif url.name IS "index">
 	<cfset data = {name="CFDocs", description="Ultra Fast CFML Documentation", type="index"}>
 <cfelseif FileExists(ExpandPath("./guides/en/#url.name#.md")) OR url.name is "how-to-contribute">
+
+	<cfset request.hasExamples = true>
```
pfreitag, line 7: Is this line supposed to be here?

scottsteinbeck: yes you have an if statement added for the css and js files so only pages with examples load the prettify library

scottsteinbeck: since the verbiage made since "hasExamples" and since guides are in markdown so we can't guarantee (without some additional parsing/checking) there are in fact no examples on the page i felt this was the simplest approach

pfreitag: ok makes sense thanks

PR #1483, opened 2022-05-04
Auto-generated guides order tags/functions alphabetically

KamasamaK:
The auto-generated guides "Adobe ColdFusion only", "Deprecated", "Discouraged", and "Lucee only" listed their tags/functions without a sensible order. They are now ordered alphabetically.

I've also added a space between the `##` and secondary heading as that is bad practice and seems to break the heading with flexmark.


pfreitag:
thanks @KamasamaK

PR #1487, opened 2022-05-04
Update spreadsheetgetcellformula.json

carlhuth:
This is a working example of adding row data to a spreadsheet, adding formulas, viewing the results and viewing the formula itself.


pfreitag:
Thanks @carlhuth - it looks like there was a JSON error in your update, I'm not seeing the error offhand, but you can run it through an online json validator and it should point it out.

BTW you don't need to wrap your code examples in cfscript it will add that automatically when necessary.

data/en/spreadsheetgetcellformula.json
```diff
@@ -25,6 +25,13 @@
...
+			"result": "Total: 57 formula: SUM(C2:C10) Total: 21600 formula: SUM(D2:D10) Total: 54 formula: SUM(E2:E10)",
+			"runnable": "1"
```
KamasamaK, line 34: This should be `"runnable": true` not `"runnable": "1"`.

carlhuth: Yes, this was from the previous example that was submitted.   I just saw this as well, thank you!

carlhuth: Oh, it was my generated result.  Thank you for pointing this out!

pfreitag: Good catch @KamasamaK, I always appreciate your input!

@carlhuth - I think you must have used the new _Add Example_ button which caused this bug in the generated json, it has been fixed already, so next time you do this you would not have the same problem.

Issue #1397, opened 2021-06-25
Document Lucee 5.3.8 Features


pfreitag:
Lucee 5.3.8.189 stable has been released, and with it there are some new functions / arguments. List here: <https://dev.lucee.org/t/new-stable-release-5-3-8-189/8484>

pfreitag:
@KamasamaK - ok thanks, I'm going to close this one then, I'd like to get a better handle on the open issues we have here in this repo. Thanks again for all your hard work on this!

PR #1498, opened 2022-05-12
Version updates

KamasamaK:
This addresses numerous changes made in the latest versions of ColdFusion and Lucee.

I went through #1366 and #1397 and I believe this resolves most of the things pending.


pfreitag:
Thanks @KamasamaK this looks great!

We are getting a fail on the tests:

```
    (!)should be in index.json if it is a tag or a function (135 ms) 
	-> Failure: querySetRow was not in index.json function index
	
    (!)should be in functions.json if it is a function (16 ms) 
	-> Failure: querySetRow was not in functions.json index
```

pfreitag:
I think you can run the utilities/indexer.cfm on your local repo to fix that.

pfreitag:
Thanks @KamasamaK

PR #1499, opened 2022-05-12
Remove saveContent function and replace region-specific links

KamasamaK:
As I was replacing region-specific Adobe doc links, I found that `saveContent` is not a function. Someone mistook the script syntax for the tag as a function.


pfreitag:
Thank You @KamasamaK

PR #1500, opened 2022-05-12
Capitalize Examples subheading

KamasamaK:
With #1497, I think the subheading for "Examples" would look better if capitalized.

As I was editing this page, I also found some spacing issues I fixed. I also `var` scoped `result` in `replaceScript`.


pfreitag:
Thanks @KamasamaK

PR #1501, opened 2022-05-28
Offer a most basic cfgrid example

carehart:


carehart:
I do realize that many are firmly against UI-based CFML elements. So why have a working example? Well, I was trying to help someone with legacy code having a very complex cfgrid example, which was producing no output at all--and I just wanted to create a very simple example to confirm whether THAT would work or not in their environment.

But the CF docs offered only db-based examples, using one of the example db's which the client did not have implemented. So we crafted a simple example and I figured I'd share it here, since there were no examples here at all. There are of course more at [the ACF docs page](https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-g-h/cfgrid.html).

pfreitag:
Thanks @carehart - I always appreciate having examples, even if a tag/function is discouraged

carehart:
Great, thanks.

PR #1511, opened 2022-06-30
updated shorthand syntax

siddiq5373:
updated shorthand syntax for ordered case-sensitive struct


siddiq5373:
updated shorthand syntax for ordered case-sensitive struct

PR #1516, opened 2022-07-25
Fix example code in cleartimezone.json

cubiclabs:
The example code had some errors in it and would not run in Lucee. The expected result was also the wrong way around.


pfreitag:
Thanks!

PR #1519, opened 2022-07-29
Add column name example

simonbingham:
This change adds a column name example to the query loop documentation.


guides/en/for.md
```diff
@@ -46,8 +46,14 @@ The above example would output `appleorangebanana`
...
 
-    for (row in query) {
-        writeOutput(query.currentRow);
+    fruits = queryNew("name", "varchar", [
+        ["apple"],
+        ["banana"],
+        ["orange"]
+    ]); 
+
+    for (item in fruits) {
```
KamasamaK, line 55: I would recommend keeping the name `row` instead of `item`.

simonbingham: Yes, I agree.

pfreitag:
Thanks @simonbingham

PR #1523, opened 2022-09-06
Milliseconds & timezone update

JamoCA:
CF2021 & Lucee support `milliseconds`. Only Lucee supports `timezone`.


pfreitag:
Thanks @JamoCA

KamasamaK:
@JamoCA How did you discover that CF2021 added the millisecond parameter? I've confirmed that it does, but Adobe doesn't seem to have documented it.

JamoCA:
I discovered this by updating a [dateHash UDF](https://gist.github.com/JamoCA/8961085) that I wrote, tested on CF2021 and then quickly scrambled as it broke everything after migrating it to a CF2016 server.

I used TryCF.com to contrast and compare whether or not an error was thrown.  I could have also performed this test using CommandBox, but it was easier to identify using TryCF.

PR #1533, opened 2022-09-30
fix member syntax

savaticus:
Member usage was wrong and raised method not found error in both 2018 and 2021. Plain .append works fine and is blessedly consistent with other data type specific append methods.


pfreitag:
Thank you @savaticus

PR #1534, opened 2022-09-30
Fix member function syntax

savaticus:
Member usage was wrong and raised method not found error in both 2018 and 2021. Plain .prepend works fine.


pfreitag:
Good catch @savaticus - thanks for fixing!

PR #1535, opened 2022-10-03
Added example of member function usage for dateAdd()

homestar9:


homestar9:
@pfreitag, it looks like the following failed:
_"Failure: dateadd.json example result is:Will return a timestamp for tomorrow but evaluated to:{ts '2022-10-04 14:46:20'}. Expected [Will return a timestamp for tomorrow] Actual [{ts '2022-10-04 14:46:20'}]"_
I believe this is because I wrote the example to use `now()`. Would you prefer I use a static date for the docs instead?

pfreitag:
Yeah @homestar9 - if you can use a static date that would be perfect. Thanks!!

PR #1546, opened 2022-10-05
Update arraycontainsnocase.json

pczarn2:
Add member function minimum required CF version label on the Member function syntax example


KamasamaK:
This change is not valid JSON.

Issue #1505, opened 2022-06-17
cfinput disabled argument is incorrect


jahschwa:
Currently `cfinput` shows `disabled` as a `boolean`. However, that's not actually the behavior. Both of these result in a disabled input field:
```
<cfinput name="foo" disabled=true>
<cfinput name="foo" disabled=false>
```

This is because ColdFusion treats this attribute the same as HTML does; its presence at all activates its effect:
https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-i/cfinput.html

So while you technically _can_ pass `true` or `false`, both values exhibit the same behavior, disabling the field. The only way to keep the field enabled is to exclude the attribute entirely:
```
<cfinput name="foo">
```

I wanted to PR this, but at a glance it looks like this isn't handled in the schema, so hoped someone could chime in on the correct approach here.

Tested on CF 2018.

PR #1571, opened 2022-10-11
CFInput - Update description with html and flash behavior #1505

Xcreen:
Update description with html and flash behavior #1505


pfreitag:
thanks @Xcreen

PR #1573, opened 2022-10-13
Update cfschedule.json

pczarn2:
Clarified when the task name attribute is required. Added example of using the list attribute.


pfreitag:
If the `task` attribute it is not **always** required, then we should also set required to false

pczarn2:
Change made. Thanks!

pfreitag:
Thanks!

PR #1576, opened 2022-10-13
Add is date object function

pczarn2:
Add new CF2016 Update 3 function. Reference: https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-in-k/isdateobject.html


pczarn2:
Found an issue

PR #1578, opened 2022-10-13
Add is date object function

pczarn2:
Add new CF2016 Update 3 function. Reference: https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-in-k/isdateobject.html


data/en/isdateobject.json
```diff
@@ -0,0 +1,15 @@
+{
+	"name":"isDateObject",
+	"type":"function",
+	"syntax":"isDateObject(value)",
+	"returns":"boolean",
+	"related":["isDate","isObject"],
+	"description":"Checks if the given value is a date / time object. The difference between isDateObject and isDate, latter returns true for date / time objects as well as date / time strings.",
+	"params": [
+		{"name":"value","description":"A value, usually the name of a variable","required":true,"default":"","type":"variableName","values":[]}
```
KamasamaK, line 9: Type is `any`. The description is erroneously referring to the name of a variable, which I've just brought up to Adobe since the description is copied from their documentation. I've also brought up the fact that the param name should be `dateobject`.

data/en/isdateobject.json
```diff
@@ -0,0 +1,15 @@
+{
+	"name":"isDateObject",
+	"type":"function",
+	"syntax":"isDateObject(value)",
+	"returns":"boolean",
+	"related":["isDate","isObject"],
+	"description":"Checks if the given value is a date / time object. The difference between isDateObject and isDate, latter returns true for date / time objects as well as date / time strings.",
```
KamasamaK, line 7: The grammar here seems off. Maybe something like "The difference between `isDateObject` and `isDate` is that the latter returns..."

data/en/isdateobject.json
```diff
@@ -0,0 +1,15 @@
...
+	"params": [
+		{"name":"value","description":"A value, usually the name of a variable","required":true,"default":"","type":"variableName","values":[]}
+	],
+	"engines": {
+		"coldfusion": {"minimum_version":"2016", "notes":"", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-in-k/isdateobject.html"}
```
KamasamaK, line 12: Based on the History section in Adobe's documentation, it seems like the minimum version is actually `2016.0.3`.

KamasamaK (changes requested):

KamasamaK:
There's no need for you to keep closing your PRs. Just add a new commit for your changes. This is the 7th PR you've made to add this function.

data/en/isdateobject.json
pczarn2: Thanks for the comments. I will make changes when I find some time.

data/en/isdateobject.json
pczarn2: Have made the change and committed.

data/en/isdateobject.json
pczarn2: Have made the change and committed.

data/en/isdateobject.json
pczarn2: Made the change and committed

pczarn2:
@KamasamaK This should now be ready to merge into the main branch. Let me know if you want more changes.

pfreitag:
Thanks @pczarn2 and also @KamasamaK for review!

KamasamaK:
Sorry I didn't respond yet. I was waiting to hear back from Adobe but after a week still nothing -- unsurprising. It turns out that their History section is wrong about the parameter being renamed. It is supposed to be `value`.

PR #1579, opened 2022-10-18
Update cfwddx.json

pczarn2:
Adding an example of how to use cfwddx to read in the contents of the scheduled tasks ColdFusion file.


pfreitag:
thanks

PR #1580, opened 2022-10-20
Update isdateobject.json

pczarn2:
Renamed dateObject param to value per KamasamaK

### DIFF
--- a/data/en/isdateobject.json
+++ b/data/en/isdateobject.json
@@ -6,10 +6,9 @@
 	"related":["isDate","isObject"],
 	"description":"Checks if the given value is a date / time object. The difference between isDateObject and isDate is that the latter returns true for date / time objects as well as date / time strings.",
 	"params": [
-		{"name":"dateObject","description":"A value, usually the name of a variable","required":true,"default":"","type":"any","values":[]}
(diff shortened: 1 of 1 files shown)


data/en/isdateobject.json
```diff
@@ -6,7 +6,7 @@
 	"related":["isDate","isObject"],
 	"description":"Checks if the given value is a date / time object. The difference between isDateObject and isDate is that the latter returns true for date / time objects as well as date / time strings.",
 	"params": [
-		{"name":"dateObject","description":"A value, usually the name of a variable","required":true,"default":"","type":"any","values":[]}
+		{"name":"value","description":"A value, usually the name of a variable","required":true,"default":"","type":"any","values":[]}
```
KamasamaK, line 9: I still say get rid of the "usually the name of a variable" part. It's misleading.

pczarn2: So, just say "A value" in the description?

KamasamaK: Up to you how you want to word it. You could say something like "A value to test whether it's a date object".

KamasamaK (approved):

PR #1583, opened 2022-10-20
Fixed #1532

### LINKED ISSUE #1532: GenerateSecretKey  does not support algorithm CFMX_COMPAT as CFDocs write
In CFDocs  generatesecretkey page  and in the point about the algorithm values the value "CFMX_COMPAT" is included, which is not correct.  Using it it gives an error  message "The CFMX_COMPAT algorithm is not supported by the Security Provider you have chosen". 

It's also confirmed in the Adobe page 

https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-e-g/generatesecretkey.html
 
where "CFMX_COMPAT" is not included in the algorithm  parameters,
and also in the paragraph "Usage" of the same page it clearly says

"You cannot use the GenerateSecretKey function to generate a key for the ColdFusion default encryption algorithm (CFMX_COMPAT..."

shawnoden:
Removed CFMX_COMPAT from algorithm values.

### DIFF
--- a/data/en/generatesecretkey.json
+++ b/data/en/generatesecretkey.json
@@ -6,7 +6,7 @@
 	"related":["encrypt","decrypt"],
 	"description":"Generates a secure random key value for use in the encrypt and decrypt functions.",
 	"params": [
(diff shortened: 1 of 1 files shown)


pfreitag:
Thanks @shawnoden

Issue #1564, opened 2022-10-07
cfchart


cfcoder:
format = "flash"? Flash is dead. Now this is a required attribute and has no default per Adobe docs.

Xcreen:
If i checked it correctly...
Acf default was flash till version 11.
For acf 11 the default seems to be png.
Since acf 2016 the default seem to be html.
In Lucee the default seems to be png.  
Also html is not supported by lucee.

Should we write the latest acf as standard or the lucee one?

pfreitag:
I would use ACF as the default, but make sure to include that for lucee the default is png

PR #1585, opened 2022-10-21
- Added html to possible values

Xcreen:
- Added html to possible values
- Changed default to html
- Added note that default for lucee is png and html is not supported

Issue: #1564


pfreitag:
Thanks @Xcreen

PR #1587, opened 2022-10-24
Added memberReturns key (structAppend)

Daemach:
Member functions can return different values.


KamasamaK:
`memberReturns` is not the proper key for CF Docs. If you check out the [JSON Schema](https://github.com/foundeo/cfdocs/blob/master/data/cfdocs.schema.json), you'll see that `member_details` is what should be used for this.

KamasamaK:
For `member_details`, the `name` and `type` are also required, which is why it is failing JSON validation. In this case, name would be `append` and type would be `struct`.

pfreitag:
Thanks!

PR #1588, opened 2022-10-26
Update onapplicationend.json

pczarn2:
Adding example of how onApplicationEnd could be used.

### DIFF
--- a/data/en/onapplicationend.json
+++ b/data/en/onapplicationend.json
@@ -14,6 +14,14 @@
 	},
 	"links": [
 
+	], 
+	"examples": [
+		{
+		"title": "Write to log when onApplicationEnd is triggered",
+		"description": "Write to the log that application has ended",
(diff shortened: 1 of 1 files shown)


pfreitag:
The cfscript tag is not necessary here, it is implied if you don't use tags in the example.

In this example, maybe we should just use cflog instead of fileWrite? Then the example is not reliant on a windows file system path and would work cross platform.

pczarn2:
@pfreitag Hi Pete, just wondering when this change and some other older changes can be merged into the documentation?

pczarn2:
@pfreitag Hi Pete, just wondering if the change I made on Oct 31 is good for merging this request.

pfreitag:
Hi @pczarn2 - sorry for the delay in getting back to you... the example is great, one other issue I'd like to fix before merging is that the `"result"` key is intended for an actual result that can be tested, not a description of what happens. Can you move what you have in the result key into the description? Sorry to be a pain, just trying to keep consistency.

pczarn2:
> Hi @pczarn2 - sorry for the delay in getting back to you... the example is great, one other issue I'd like to fix before merging is that the `"result"` key is intended for an actual result that can be tested, not a description of what happens. Can you move what you have in the result key into the description? Sorry to be a pain, just trying to keep consistency.

@pfreitag Sure, I'll work on that when I get a chance.

pczarn2:
@pfreitag Just letting you know I moved the "result" description to "description" field.

pfreitag:
Thanks @pczarn2

PR #1594, opened 2022-11-01
Update cfrethrow.json

scottsteinbeck:
add file read example with rethrow


pfreitag:
Thanks @scottsteinbeck - again cfscript tag not needed for the example

PR #1595, opened 2022-11-01
Update isdateobject.json

scottsteinbeck:


pfreitag:
Thanks @scottsteinbeck - The `<cfscript>` tag is not necessary in the example here it is implied if omitted.

PR #1596, opened 2022-11-01
Update toscript.json

scottsteinbeck:


pfreitag:
Example is a little long, prefer shorter examples but I can see how it might be useful here.

PR #1600, opened 2022-11-02
Update isvalid.json

pczarn2:
Added "time" to list of "type" attribute values. Moved couple atributes values to be listed in alpha order. min, max and pattern attributes are not required, they are only required for specific cases, so they should be listed as optional.


pfreitag:
thanks!

PR #1601, opened 2022-11-02
Update listgetat.json

pczarn2:
Corrected argument name to delimiter. This can be checked by calling people = "Mick,,,,Keith,,,,Charlie,Ronnie,,,,,Bill"; WriteOutput( "The second item is: #ListGetAt( list = people, position = 2, delimiter="," )#" ); If called with "delimiters" param instead, it throws an error (tested on CF2018).


KamasamaK:
Good catch. I just had Adobe update their official documentation too.

PR #1604, opened 2022-11-09
Fix cfimage captcha example.

faxi05:
The captcha sample code for cfimage was missing the required "width" and "height" attributes.


pfreitag:
thanks!

PR #1610, opened 2022-11-18
Update int.json

pczarn2:
Added note about discouraged use of this function. Source: https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-in-k/int.html

### DIFF
--- a/data/en/int.json
+++ b/data/en/int.json
@@ -5,6 +5,7 @@
 	"returns":"numeric",
 	"related":["round","floor","ceiling"],
 	"description":" Calculates the closest integer that is smaller than number.",
+	"discouraged": "In new applications, use the Floor function instead of the Int function.",
 	"params": [
 		{"name":"number","description":"","required":true,"default":"","type":"numeric","values":[]}
 	],


pczarn2:
@pfreitag I do not know the reason, all I saw on their page is a note that they discourage using it.

KamasamaK:
I think instead of "new applications", it should specify CF2016+.

pfreitag:
Thanks @pczarn2 - I am going to post this on the CFML slack to see if anyone has any thoughts on it. I don't want to mark something discouraged without having a good reason for it.

pczarn2:
@pfreitag I found below article today. Does this give a good reason? Would this be a good description?  

"discouraged": "In new applications, use the Floor function instead of the Int function. int() should just return the integer part of a number, but instead it performs a rounding exercise instead. This means with negative numbers, it gets the answer wrong (https://blog.adamcameron.me/2016/02/coldfusion-2016-floor.html)."

pczarn2:
@pfreitag And just a note I tried cffildle code below (reference: https://blog.adamcameron.me/2015/08/cfml-how-should-int-behave.html), results are -38 for int() (incorrect answer) and correct answer -37 for fix(), so maybe the description should be:

"discouraged": "In new applications, use the **Fix** function instead of the Int function. int() should just return the integer part of a number, but instead it performs a rounding exercise instead. This means with negative numbers, it gets the answer wrong (https://blog.adamcameron.me/2016/02/coldfusion-2016-floor.html)."

<cfscript>
writeOutput(#int(35 - (1013 / (13+1)))#)
</cfscript>

pfreitag:
I am going to close this one, I don't feel that `int()` should be marked as _discouraged_ because I think _discouraged_ should be for places where the developer should really consider rewriting the code. In the case of int, it is doing exactly what it is documented to do, `Calculates the closest integer that is smaller than number.` so `int(-3.4)` correctly returns `-4`. Yes, there is now a better named alias `floor`, and yes it _should_ be used instead since it is more clear what it is doing, but if you are already using `int()` I think that is fine. 

I feel that adding _discouraged_ is a strong warning, and I don't feel this case merits such a strong warning. Adobe / Lucee will never remove the `int()` function, there are mountains of code that use it. 

If people feel I am way off base here, I am happy to re-open and reconsider.

PR #1611, opened 2022-11-18
Update discouraged.md

pczarn2:
Added int to discouraged functions.

### DIFF
--- a/guides/en/discouraged.md
+++ b/guides/en/discouraged.md
@@ -102,6 +102,9 @@ The use of tags generating UI is generally discouraged by the CFML community. Se
 
 In almost all cases evaluate is unnecessary (use bracket notation instead). Use of evaluate can lead to performance and security issues.
 
+`int`
+Adobe recommends that you use the Floor function, instead of the Int function, in all new applications.See: https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-in-k/int.html
+
 `iIf`
 
 Use the ternary operator CF9+ instead because it does not evaluate.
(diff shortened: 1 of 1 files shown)


guides/en/discouraged.md
```diff
@@ -102,9 +102,12 @@ The use of tags generating UI is generally discouraged by the CFML community. Se
 
 In almost all cases evaluate is unnecessary (use bracket notation instead). Use of evaluate can lead to performance and security issues.
 
+`int`
+Adobe recommends that you use the Floor function, instead of the Int function, in all new applications.See: https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-in-k/int.html
```
KamasamaK, line 106: Missing a space after the period

pfreitag:
@pczarn2 two things on this: 

1) Do you know why Adobe recommends using floor instead of int? It would be useful to know the reason, for example does it support larger numbers than int? 

2) The discouraged markdown is generated automatically by adding a discouraged flag on the function or tag, rather than editing this file.

pfreitag:
Ah - I see you have done this in #1610

KamasamaK:
@pczarn2 If you're autogenerating this using `/utilities/guide.cfm` then this shouldn't require a manual review. I think this is just waiting on #1610.

zspitzer:
As a heads up, Lucee already internally just uses `Math.floor()` for `int()`

https://github.com/lucee/Lucee/blob/5.3/core/src/main/java/lucee/runtime/functions/math/Int.java#L36

pczarn2:
Closing, since the discouraged markdown is generated automatically by adding a discouraged flag on the function or tag, rather than editing this file.

PR #1615, opened 2023-01-23
Update cfmail.json

dbelanger:
This example adds the syntax for Lucee which is weird because it's not actually a function call and tripped me up a few times.


pfreitag:
thanks @dbelanger

PR #1623, opened 2023-04-24
Added "result" attribute to cflock for Lucee

scallahanCybernet:
Lucee has a result attribute for returning the statusCode and ExcecutionTime of the lock as a variable.


pfreitag:
thanks @scallahanCybernet

PR #1624, opened 2023-04-26
EntityLoadByPK() does not have a `unique` parameter in Lucee

michaelborn:
This PR drops the `unique` attribute from the `EntityLoadByPK()` docs, since it hasn't existed in the Lucee source code since 2015.

Maybe this was 4.5 only? Or did it ever exist at all?

https://github.com/lucee/Lucee/blob/6.0/core/src/main/java/lucee/runtime/functions/orm/EntityLoadByPK.java#L11


pfreitag:
thanks @michaelborn

PR #1629, opened 2023-09-08
Use text() instead of html()

shaedrich:
To prevent XSS (even though, unlikely) and improve performance (even though, only slightly)


pfreitag:
yes - I'm a big fan of using innerText instead of innerHTML - thanks for the PR @shaedrich

shaedrich:
Phew! I was worried it was too little of a change 😅 

[`textContent`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) is [even better](http://perfectionkills.com/the-poor-misunderstood-innerText/)

PR #1631, opened 2023-10-03
Fix return type of writeOutput

jimpriest:
Noticed the return type for writeOutput was incorrect.  The Adobe docs don't even mention it but the Lucee docs are correct: https://docs.lucee.org/reference/functions/writeoutput.html


pfreitag:
Thanks!

PR #1632, opened 2023-10-10
Update monthasstring.json

mubbo:
Added missing optional parameter "Locale", added example using optional parameter

### DIFF
--- a/data/en/monthasstring.json
+++ b/data/en/monthasstring.json
@@ -1,7 +1,7 @@
 {
 	"name":"monthAsString",
 	"type":"function",
-	"syntax":"monthAsString(month_number [, Locale])",
+	"syntax":"monthAsString(month_number [, locale])",
 	"returns":"string",
 	"related":["month","monthShortAsString"],
 	"description":" ",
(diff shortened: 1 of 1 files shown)


data/en/monthasstring.json
```diff
@@ -1,12 +1,13 @@
 {
 	"name":"monthAsString",
 	"type":"function",
-	"syntax":"monthAsString(month_number)",
+	"syntax":"monthAsString(month_number [, Locale])",
```
pfreitag, line 4: Can we change this to a lower case `L`?

pfreitag:
@mubbo - thanks for the fix, looks like there is a JSON error in the file, can you check that: https://github.com/foundeo/cfdocs/actions/runs/6471910790/job/17572582747?pr=1632#step:7:712 

Once it passes validation I can merge it - thanks!

data/en/monthasstring.json
```diff
@@ -21,6 +22,12 @@
...
 			"code": "getVal = monthAsString(11); \r\n writeDump(getVal);",
 			"result": "result is: November"
-		}
+		},
+		{
+			"title": "Script Syntax",
+			"description": "With locale",
+			"code": "getVal = monthAsString(6,\"DE\"); \r\n writeDump(getVal);",
+			"result": "result is: Juni"
+		},
```
pfreitag, line 31: extra comma on line 31

PR #1633, opened 2023-11-09
Update cfcontent.json

bendur:
Changed existing example to not runnable, added a cfscript example.


pfreitag:
Thank you @bendur !!

PR #1634, opened 2023-12-17
Run once utility to automate creation of "member_details" where it doesn't already exist

garethedwards-tass:


pfreitag:
@garethedwards-tass - in general I'm ok with adding the member_details, but as for the utility is there a reason to use custom built json functions, rather than the builtin: `isJSON`, `serializeJSON` and `deserializeJSON`?

garethedwards-tass:
For these changes in particular the goal was to reduce additional changes to the json files

KamasamaK:
`base_param_index` should be a number, not a string.

ghedwards:
@KamasamaK I've removed the utilities

ghedwards:
@KamasamaK I've updated the return types

garethedwards-tass:
Maybe I'll try another approach ?

PR #1636, opened 2023-12-27
Added "function" as possible type-parameter #1628 

### LINKED ISSUE #1628: isValid
isValid type option:

Docs should be updated to reflect that function/closure is an option for 'type'. 
I don't know if this is undocumented but it works in ACF and Lucee.

**Function** works in ACF and Lucee:
`isValid( "function", function(a,b) {return a+b;} );`

**Closure** only works for Lucee:
`isValid( "closure", function(a,b) {return a+b;} );`

Xcreen:
- Added "function" as possible type-parameter
- Lowercase all type-parameters
- Sort all type-parameters

close #1628


pfreitag:
thanks

PR #1637, opened 2024-02-09
Add nullvalue example

tomchiverton:


pfreitag:
@tomchiverton - Have you tested this one, I don't think that will work

tomchiverton:
IDK where the # went, it should be

	"code": "queryExecute(\"\n    update foo set\n    staff_group_id = #isDefined('form.staff_group')?form.staff_group:nullValue()#\n    where bar='baz'\n\")",

PR #1639, opened 2024-02-16
Add member function to datecompare.json

scallahanCybernet:
Added the member function to datecompare.json as well as an example.


pfreitag:
Thanks

PR #1640, opened 2024-02-29
Corrected isLeapYear example to pass a numeric year instead of a date…

bockensm:
Both Lucee & Adobe want a year passed to the method, not a date object.


pfreitag:
Quite the day for such a change - thanks @bockensm

PR #1641, opened 2024-03-01
Use static date in example

bockensm:
The first example fails to match the expected result in a leap year, so I switched it to a static date on a non-leap year.


pfreitag:
Thanks!

PR #1642, opened 2024-03-06
Update serialization options

KamasamaK:
- Updated parameter names for `serialize` and `serializeJSON` per #898
- Fixed some other parameter information
- Described some serialization changes made in CF11
- Added example for serializing a query to JSON
- Improved example for serializing to XML


pfreitag:
Wonderful updates - thanks @KamasamaK

PR #1643, opened 2024-03-08
Update for CF2023

KamasamaK:
- Added new CF2023 functions
- Added CF2023 to TryCF engines and made the new default

Note: `generateGraphQLModels` and `getGraphQLClient` do not have dedicated function pages on Adobe's ColdFusion help site, so I've used their GraphQL article as the "docs" link for now.


KamasamaK:
Even though I ran the utilities, I still got an error for the new `cf2023.json` not being present in the `index.json`. I found that this is because those files are generated after checking for what to include in the `index.json`, so I moved that part earlier in the process. There might be similar issues with the other files created too late as well, but I didn't want to make and changes beyond the scope of this particular issue. Running the utility again would add them anyway, but that should not be necessary.

KamasamaK:
Even though I ran the utilities, I still got an error for the new `cf2023.json` not being present in the `index.json`. I found that this is because those files are generated after checking for what to include in the `index.json`, so I made a fix for that. Running the utility again would add them anyway, but that should not be necessary.

pfreitag:
Amazing work @KamasamaK thank you!

PR #1644, opened 2024-03-13
Related: Add cfzipparam

JamoCA:
cfzipparam referenced cfzip, but cfzip didn't reference cfzipparam


pfreitag:
The circle is now complete. Thank you @JamoCA

PR #1645, opened 2024-03-26
Update arraynew.json

pczarn2:
Correct default value for isSynchronized argument; remove no longer working link to Adobe white paper; correct Create unsynchronized array example. Reference: https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-a-b/arraynew.html

### DIFF
--- a/data/en/arraynew.json
+++ b/data/en/arraynew.json
@@ -7,20 +7,13 @@
 	"description":"Creates an array of 1-3 dimensions. Index array elements with square brackets: [ ]. CFML arrays expand dynamically as data is added.",
 	"params": [
 		{"name":"dimension","description":"","required":true,"default":1,"type":"numeric","values":[1,2,3]},
-		{"name":"isSynchronized","description":"CF2016+ Lucee5.1+ When true creates a synchronized array. Unsynchronized arrays are not thread safe so they should not be used within shared scopes (application, session, etc). According to the CF2016 Performance whitepaper: Unsynchronized arrays are about 93% faster due to lock avoidance.","required":false,"default":false,"type":"boolean","values":[]}
+		{"name":"isSynchronized","description":"CF2016+ Lucee5.1+ When true creates a synchronized array. Unsynchronized arrays are not thread safe so they should not be used within shared scopes (application, session, etc). According to the CF2016 Performance whitepaper: Unsynchronized arrays are about 93% faster due to lock avoidance.","required":false,"default":true,"type":"boolean","values":[]}
 	],
 	"engines": {
 		"coldfusion": {"minimum_version":"3", "notes":"", "docs":"https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-a-b/arraynew.html"},
 		"lucee": {"minimum_version":"", "notes":"", "docs":"https://docs.lucee.org/reference/functions/arraynew.html"},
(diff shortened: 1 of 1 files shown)


pfreitag:
@pczarn2 - did you check to see if there is another link to that PDF, perhaps archive.org may have it?

pczarn2:
@pfreitag Adobe's own website has the link, but that link is broken (https://helpx.adobe.com/coldfusion/white-papers.html). Archive.org has the webpage, but the link to PDF is broken there as well. There were other pages that have the same broken link. I did find some working links here, but it's in Japanese, so not too useful to me :) Not sure where else to check. 

https://www.google.com/search?q=CF2016_Performance.pdf&oq=CF2016_Performance.pdf&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIGCAEQRRg8MgYIAhBFGDzSAQc3MThqMGoxqAIAsAIA&sourceid=chrome&ie=UTF-8

KamasamaK:
The PDF is on archive.org:
https://web.archive.org/web/20211021051748/http://wwwimages.adobe.com/content/dam/acom/en/products/coldfusion/pdfs/cf2016/CF2016_Performance.pdf

zspitzer:
We should be careful about citing performance stats, whilst it makes theoretical sense, it seems to be way overstated?

https://trycf.com/gist/7e7c04ef1614adf9bea423f67d0e007a/lucee5?theme=monokai

I see little difference in this little microtest with 50k appends? also seems the named parameters don't work in ACF 2023? (there's also acf uses dimensions, lucee has dimension)

pczarn2:
@KamasamaK Thanks for finding the reference!

@zspitzer without named params, this code works with ColdFusion engines:
<cfscript>
  a = arrayNew(1,false);
    // Implicit array notation
    a.append([]);
    // with values
    a.append( [
        'a','b',
        3, 4,
        [],
        {complex: true},
        queryNew("id,date")
    ] );
    writedump(a);
</cfscript>

pczarn2:
@pfreitag Pete, I added back the link with the archived version @KamasamaK found. Can you merge the change?

KamasamaK:
@zspitzer 
> seems the named parameters don't work in ACF 2023? (there's also acf uses dimensions, lucee has dimension)

I am seeing that named parameters are working fine in ACF 2023. You are correct about the name difference with Lucee. Seems like the addition of an alias would be the way to go there.


@pczarn2 
> without named params, this code works with ColdFusion engines: ...

I am surprised that works since Lucee has `type` as its second parameter. I wonder if it has a compatibility consideration where if it detects a Boolean value then it uses the signature with `synchronized` as the second parameter.

zspitzer:
@KamasamaK  https://luceeserver.atlassian.net/browse/LDEV-4822 :)

pfreitag:
Merged the PR thanks @pczarn2! 

Though I am now wondering if the performance is quite overstated based on what @zspitzer is saying but also on an article I found where @aliaspooryorik tested the feature, and didn't see much difference: https://coldfusion.adobe.com/2019/03/look-unsynchronised-arrays-cfml/ 

At the very least we should probably add an _allegedly_ to the performance claims.

zspitzer:
Aka, here's a useless rabbit hole, enjoy!

aliaspooryorik:
Yeah - I did ask how they got that 90% performance increase but didn't hear back so none the wiser I'm afraid!

PR #1647, opened 2024-03-29
Update arraynew.json

pczarn2:
Added word allegedly to CF2016 White Paper performance claims


pfreitag:
Nice! Thanks @pczarn2

PR #1648, opened 2024-03-31
Update CF min version and more (A-O functions)

KamasamaK:
I had to stop at O to keep it more manageable.

The goal was to update the CF minimum versions, but I also made various other changes for those files that had other issues such as

- Formatting
- Param names
- Missing parameters
- Missing or incorrect descriptions

The exception is `arrayGetMetadata` to move an incorrect example from `getMetadata`.


KamasamaK:
I also made some tweaks to the "ColdFusion Member Functions" guide as I was adding the new `millisecond` item to it.

Issue #1027, opened 2018-10-26
Add a .prettierrc (or .prettierrc.json) to auto-format JSON (and JS) files


atuttle:
[Prettier](https://github.com/prettier/prettier#readme) is great. Having a definitions file will ensure that (at least for people who have the extension installed in their editor), the files are all consistently formatted.

pfreitag:
Good idea -- if you want to submit a PR we can see how well it might work.

atuttle:
Here's the config I would recommend, but it does incorporate some of my own personal preferences.
```
{
  "arrowParens": "avoid",
  "bracketSpacing": true,
  "htmlWhitespaceSensitivity": "css",
  "insertPragma": false,
  "jsxBracketSameLine": false,
  "jsxSingleQuote": false,
  "parser": "json",
  "printWidth": 80,
  "proseWrap": "preserve",
  "requirePragma": false,
  "semi": true,
  "singleQuote": false,
  "tabWidth": 4,
  "trailingComma": "none",
  "useTabs": true
}
```

[...]

PR #1649, opened 2024-04-20
Feature/prettier cflint

psarin:
Adding prettier and cflint rules as per https://github.com/foundeo/cfdocs/issues/1027. Upon agreement of rules, will refort JSON and other files based on rules.


pfreitag:
Thanks for starting this - for using prettier on the json files I think it is a good idea, can you confirm that the rules you have created are best matching for the format we currently have?

PR #1650, opened 2024-04-25
OpenBD is dead, so stop linking to it

tomchiverton:
No code changes in >4 years, unattended security issues such as https://github.com/OpenBD/openbd-core/issues/89 (opened in Sept 2023, no replies)


pfreitag:
Yes, good idea - thanks

PR #1651, opened 2024-04-28
Update right.json

aliaspooryorik:
right now supports negative counts


pfreitag:
wow didn't know that - thanks @aliaspooryorik 

If you know what version that was added we could add engine notes for that.

aliaspooryorik:
Hi @pfreitag - it was added in the ACF 2018 release. I probably should update `left` as well!

PR #1652, opened 2024-04-29
Update queryappend.json

scallahanCybernet:
Added support for Lucee 6.0.0.15

### DIFF
--- a/data/en/queryappend.json
+++ b/data/en/queryappend.json
@@ -2,7 +2,7 @@
 	"name":"queryAppend",
 	"type":"function",
 	"syntax":"queryAppend(query1, query2)",
-    "member":"qry.append(query2)",
+    	"member":"qry.append(query2)",
 	"returns":"query",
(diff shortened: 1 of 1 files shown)


scallahanCybernet:
@pfreitag You may want to update the Lucee version regex in the JSON validator to allow for micro versioning. This didn't pass JSON validation when I entered `6.0.0.15`, but that is the official version this function was added to Lucee.

pfreitag:
yeah, I think we were hoping to support semver style versions there. Could you add in the notes of the engine key the exact version?

pfreitag:
going to merge this since the number of people running a version of 6.0.0 less than 6.0.0.15 has to be close to zero, but feel free to add the note in another PR

scallahanCybernet:
> going to merge this since the number of people running a version of 6.0.0 less than 6.0.0.15 has to be close to zero, but feel free to add the note in another PR

Unless Brad is still messing around with testing something 😝

PR #1658, opened 2024-06-23
Update datediff.json

penhorwood:
The current example is just code but doesn't run.  My example runs.
Added a working example


pfreitag:
Thanks!

PR #1660, opened 2024-07-11
Update csrfverifytoken.json

burncitiesburn:
csrfVerifyToken only checks the first 40 characters of the token passed in.

This should be included in the documentation as it means something like

csrfToken = 'ABC123ABCABC123ABCABC123ABCABC123ABCDEFG'&'123'

Will still verify if the generated token from generateCSRFToken is 'ABC123ABCABC123ABCABC123ABCABC123ABCDEFG'

Screenshot 2024-07-11 at 2 32 44 PM

### DIFF
--- a/data/en/csrfverifytoken.json
+++ b/data/en/csrfverifytoken.json
@@ -6,7 +6,7 @@
 	"related":["csrfGenerateToken"],
 	"description":"Validates the passed in token against the token stored in the session for a specific key. Used to help prevent Cross-Site Request Forgery (CSRF) attacks.",
 	"params": [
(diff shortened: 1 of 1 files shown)


pfreitag:
@burncitiesburn good find, this appears to only apply to Adobe ColdFusion, and not other engines such as Lucee, so it would be worth noting that as well.

burncitiesburn:
@pfreitag I've updated it to read "For Adobe Coldfusion, only the first 40 characters of the passed in token are used to verify."

pfreitag:
Thanks!

Issue #1665, opened 2024-09-13
Missing Lucee 6 Function configImport


pfreitag:
https://docs.lucee.org/reference/functions/configimport.html

cybersonic:
Hopefully we can close this issue @pfreitag ?

pfreitag:
Yup, thanks @cybersonic

PR #1674, opened 2024-10-11
#1665 adding a configImport example

cybersonic:
Adding docs for configImport see: 
Hopefully, this will help as part of #hacktoberfest (see #1665)


pfreitag:
@cybersonic thanks! - failed the json schema validation due to the returntype value: data/returns must be equal to one of the allowed values

I think you just need to make it lower case `Struct` to `struct`

pfreitag:
Thanks @cybersonic

PR #1676, opened 2024-10-19
Update cfcatch.json

JCOvergaar:
Added CFCATCH variables available when an exception is caught, copied from https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-c/cfcatch.html


pfreitag:
Nice enhancement thanks @JCOvergaar 

It looks like you accidentially chopped off the end of the param, so it is missing:

```
,"required":false,"default":"any","type":"string","values":["application","database","template","security","object","missinginclude","expression","lock","custom_type","searchengine","any"]```

Could you add that back in?

Issue #1521, opened 2022-08-05
Missing Lucee's IsInThread method


michaelborn:
## IsInThread

> Determines whether the executing code is inside a cfthread or not.

(This appears to be a Lucee-only method.)

https://docs.lucee.org/reference/functions/isinthread.html

D-Anir:
Hi, can you assign this issue to me if it is still open?

pfreitag:
go for it

cybersonic:
Added a Pull Request for this issue: https://github.com/foundeo/cfdocs/pull/1678

PR #1678, opened 2024-10-21
#1521 isinthread

cybersonic:


pfreitag:
Thanks @cybersonic 

fixes #1521

Issue #1306, opened 2020-04-20
New Lucee Function: queryLazy


pfreitag:
New in Lucee 5.3.6.45: https://luceeserver.atlassian.net/browse/LDEV-2835 

    queryLazy(sql, listener, params, options);

The `listener` is a function that is passed the recordset.

PR #1679, opened 2024-10-21
#1306 Adding querylazy docs

cybersonic:
#1306  Adding docs for queryLazy


pfreitag:
Thanks @cybersonic - would be great to get the version it was added to Lucee in there too if you happen to know it.

PR #1681, opened 2024-10-30
Added member function to querysetcell.json

scallahanCybernet:
Added member function and example to querysetcell.json


pfreitag:
Thanks @scallahanCybernet

PR #1684, opened 2025-01-14
Drop param from ormcloseallsessions.json, there is no param in ACF or Lucee

michaelborn:
See ACF docs: https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-m-r/ormcloseallsessions.html

and Lucee docs: https://docs.lucee.org/reference/functions/ormcloseallsessions.html

It is clear this BIF does not take arguments.


pfreitag:
Thanks @michaelborn

PR #1685, opened 2025-01-15
var scope errors on Adobe ColdFusion in examples for querysome & structsome

akashbavlecha:
The code examples in the documentation for 

querysome (https://cfdocs.org/querysome) and 
structsome (https://cfdocs.org/structsome) utilise var scoping that causes execution failures in the default Adobe ColdFusion engine when the "Run Code" button is clicked.  My solution ensures compatibility with both CFML and Lucee.

Issue: 
image


pfreitag:
thanks!

PR #1686, opened 2025-01-20
Update arraysort.json

scallahan-dev:
Added example for script member function to sort an array of structs by more than one key.


pfreitag:
Thanks @scallahan-dev

PR #1689, opened 2025-02-20
surface search on narrow screens

atuttle:
> [!CAUTION]
> I was not able to get cfdocs running locally, so I have not been able to test these changes. However, they are small and I believe they should work. I did experiment with browser devtools to verify approaches.

The CFDocs site is responsive, which is great. But I would imagine that the majority use-case is to use the search, and when visiting on a narrow screen the search is hidden away inside the hamburger menu.

This change would -- only on narrow screens -- hide the form that collapses into the hamburger menu, and display a copy of the search form just underneath the site masthead.

### DIFF
--- a/assets/script.js
+++ b/assets/script.js
@@ -62,6 +62,7 @@ $(document).ready(function() {
 	  });
 	  $('.tt-hint').addClass('form-control');
 	  $('#search').submit(submitSearch);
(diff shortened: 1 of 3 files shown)


pfreitag:
Thanks @atuttle

PR #1690, opened 2025-02-27
Update coldfusion-versions.md

JamoCA:
Added all bold title & highlights from https://helpx.adobe.com/coldfusion/using/whats-new.html


JamoCA:
After committing this, I'm not sure if the following changes should be referenced or not:
- CF2025 now runs on Java 21 by default
- Installed with Tomcat 10.1

pfreitag:
Thanks @JamoCA

PR #1695, opened 2025-04-02
Update cfparam.json

dreamstarter:


pfreitag:
This example doesn't actually use `type="query"` on the cfparam tag. Perhaps something like this would be good:

```<cfparam name="myQuery" type="query" default="#queryNew("id")#">```

Or if you wanted a more complex example like you have you could use a cftry/cfcatch

pfreitag:
closing due to invliad json, feel free to fix and I will reopen / merge it

PR #1696, opened 2025-04-03
Update cfwhile.json

scallahan-dev:
Added support for ColdFusion for the script syntax as well as a script syntax example.

### DIFF
--- a/data/en/cfwhile.json
+++ b/data/en/cfwhile.json
@@ -10,7 +10,8 @@
 		{"name":"label","description":"Used to address loop from `break` or `continue` statement instead of innermost loop.","required":false,"default":"","type":"string","values":[]}
 	],
 	"engines": {
-		"lucee": {"minimum_version":"", "notes":"", "docs":"https://docs.lucee.org/reference/tags/while.html"}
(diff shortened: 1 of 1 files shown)


pfreitag:
Hi @scallahan-dev thanks for your contribution. I would prefer to keep the coldfusion key out of the engines struct in this case since the tag is not supported on CF the omission implies this and it will display a message. 

Would love to have your cfscript while loop example in there though, so if you want to update the PR to include just that, I'd gladly accept that.

scallahan-dev:
@pfreitag That makes since, but the notice on the page is confusing and doesn't indicate anywhere that the script syntax is valid for ColdFusion. I'm open to ideas on how to make this more apparent.

pfreitag:
Yes, good point -- maybe in your example you can mention in the description of your example that the script syntax works on both Adobe ColdFusion and Lucee

scallahan-dev:
@pfreitag New commit with your requested changes.

scallahan-dev:
@pfreitag Just wanted to see if there were any other changes you wanted made to this PR.

pfreitag:
Thanks

PR #1699, opened 2025-04-11
Update for Lucee 6.2 breaking change to log levels

tomchiverton:
https://dev.lucee.org/t/application-log-defaults-to-error-in-6-2/14773

### DIFF
--- a/data/en/cflog.json
+++ b/data/en/cflog.json
@@ -3,7 +3,7 @@
 	"type":"tag",
 	"syntax":"<cflog text=\"\">",
 	"related":["writeLog"],
-	"description":"Writes a message to a log file.",
+	"description":"Writes a message to a log file. Note in Lucee 6.2 or later the default log level is ERROR, so no messages will appear by default.",
 	"params": [
 		{"name":"text","description":"Message text to log.","required":true,"default":"","type":"string","values":[]},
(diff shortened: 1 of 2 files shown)


pfreitag:
@tomchiverton thanks for this - I was a little confused when I first read your update which says "Note in Lucee 6.2 or later the default log level is ERROR, so no messages will appear by default."

I had thought this meant the default type for cflog was changed to `error`, but you mean that the lucee appender only writes logs to file with type=error or type=fatal. 

So a few suggested updates. 

1. Instead of putting this in the description, let's move it to the lucee engine notes
2. Let's clarify this a bit, maybe something like this: _"In Lucee6.2+ the default appender log level is ERROR, so only logs with type error or fatal would be written to the log file by default"_

tomchiverton:
How do I add an engine specific note ?

carehart:
Looks like doing that last step fell between the cracks. @tomchiverton, would you want to do it (to close out your pr) or would you want one of us to proceed with a new one?

BTW, guys, I don't "live" in github. If a new or isn't necessary (for someone else to make the change), feel free to clarify. I'm just trying to help, not playing traffic cop. :-) I came on this while reviewing the Lucee breaking changes docs. 

Also, it seems it could be helpful to mention in the note that the new default log level for Lucee CAN be changed (rather than leaving the implication that every cflog/write log must be changed). That discussed here: https://dev.lucee.org/t/application-log-defaults-to-error-in-6-2/14773

tomchiverton:
I've spent the last half hour fighting Microsoft's website to let me edit two files, with 50% success so far :/

tomchiverton:
I think GitHub's AI infested git client has let me make the required four line change now ?

PR #1702, opened 2025-04-23
REEscape add Lucee 

zspitzer:
Does minimum_version support values like "6.2" i.e. when this was introduced?


pfreitag:
Yes, 6.2 could be used there instead

pfreitag:
Thanks @zspitzer!

PR #1705, opened 2025-05-02
queryrowdata.json - change calls in examples

fmancardi:
In the examples instead of queryRowData() and rowData()  queryGetRow() and getRow() where used.


pfreitag:
thanks @fmancardi

PR #1706, opened 2025-06-04
Link to `isValid` from `cfparam`

JoelNiemela:
Links to `isValid` from `cfparm`, as `cfparam` uses the same type of validation as `isValid` does (there is already a link from `isValid` to `cfparam`)


pfreitag:
good one, thanks @JoelNiemela

PR #1709, opened 2025-06-05
Add example to `isSimpleValue` for list and null

JoelNiemela:
As both `list` and `null` may be surprising edge cases for some.


pfreitag:
Thanks @JoelNiemela

PR #1713, opened 2025-07-01
Generate Markdown Docs

ghedwards:


pfreitag:
@ghedwards - can you provide some background for the reason for doing this? I assume it has something to do with making it easier for LLM's to read the docs. Is there a reason for putting them in the /docs sub folder? the guides are already written in markdown so we are just moving those files.

ghedwards:
@pfreitag context7 LLM / MCP but also cfmleditor.github.io/cfdocs

ghedwards:
Without markdown docs

image

With markdown docs

image

ghedwards:
@pfreitag if the generation is on the table why not individual .MD files, it also reduces the reliance and traffic to the cfdocs website, from llms at least.

garethedwards-tass:
@pfreitag any update on this ?

PR #1715, opened 2025-07-10
Add script syntax for cfrethrow

carehart:
(I didn't add an engine value for boxlang "while I was at it", as their docs don't currently indicate support: https://boxlang.ortusbooks.com/boxlang-language/reference/components/system/rethrow)


pfreitag:
Thanks @carehart

PR #1723, opened 2025-09-26
Update script.js

NCowoski:
The search field on full sized screens has been broken for quite some time.

The changes I added will allow both to be used together on the index.cfm and on all other pages as well.

I did kinda  throw this together. Any suggestions would be appreciated if declined.


pfreitag:
ah I see, so we probably should be using a class instead of an id then if there are two elements with the same id - I will merge this for now if you want to make it better feel free

PR #1726, opened 2025-12-15
Added examples for using more advanced Lucee loops

tom-clik:
Updated some examples for using

1. key, value in loops over structs
2. index, value in loops over arrays


pfreitag:
Thanks @tom-clik

PR #1728, opened 2026-01-22
Add member field to structCount JSON definition

scallahan-dev:


pfreitag:
Thanks @scallahan-dev

PR #1729, opened 2026-01-22
Add member field to structClear JSON definition

scallahan-dev:


pfreitag:
Thanks @scallahan-dev sorry for the delay in merging!

PR #1730, opened 2026-01-22
Update structEquals JSON with member function and example

scallahan-dev:


pfreitag:
@scallahan-dev thanks!

PR #1731, opened 2026-04-28
Add Name parameter for CFCatch

JCOvergaar:
Added name parameter -- see https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-tags/tags-c/cfcatch.html


pfreitag:
Thanks for contributing @JCOvergaar there was an extra comma in the json that I removed.

BTW - do you know what version CF added this attribute? It would be good to add that if we can figure it out. I assume it was not _always_ supported since we didn't have it yet.

JCOvergaar:
That's what I get for editing code on my phone 😔

It's in the docs for [CF10](https://help.adobe.com/en_US/ColdFusion/10.0/CFMLRef/coldfusion_10_cfmlref.pdf), and not in the docs for 9, so I'll guess added in 10 even if they didn't mention it in the change list. :)

